PR #15, opened 2025-01-07
Rename aranya spec doc name

smckeon-so:
This PR renames the Aranya Beta v1 document to more closely resemble our naming conventions as this doc includes references to the beta and MVP specifications.


gknopf-aranya (approved):

gknopf-aranya (changes requested):
`aranya-beta-v1.md`

gknopf-aranya (approved):

ycarmy (commented):
I don't think it makes sense to do this. The spec in this doc is for the entire beta version of the product - i.e., it includes all content related to both the V1 and V2 releases. Thus, the filename should be kept as `aranya-beta` IMO.

gknopf-aranya:
> I don't think it makes sense to do this. The spec in this doc is for the entire beta version of the product - i.e., it includes all content related to both the V1 and V2 releases. Thus, the filename should be kept as `aranya-beta` IMO.

I thought the main differences between V1 and V2 were the C API and AFC (which has its own spec). So the beta V1 spec didn't really change when we completed V2. I don't recall us updating the spec after V2 so it could be out-of-date.

ycarmy:
> I thought the main differences between V1 and V2 were the C API and AFC (which has its own spec). So the beta V1 spec didn't really change when we completed V2. I don't recall us updating the spec after V2 so it could be out-of-date.

Take a look at the spec. It includes mentions of both the C API and AFC. Further, it includes the interface details for AFC, as intended by this spec.

smckeon-so:
Okay, thanks we'll keep it as is.

Issue #5, opened 2025-01-06
open-source the aranya product architecture


gknopf-aranya:
Open source the Aranya product architecture doc.
Currently it's a Visio doc:
https://spideroakmscom-my.sharepoint.com/:u:/g/personal/gknopf_spideroak_com/ETeUyhYUcBRPhihbth8a63IBAUPgHG3xlwQ8XLImyBXDAw?e=KXjRBM

cc: @ycarmy, @sericso, @nikki-aranya

gknopf-aranya:
@ycarmy I'm assigning this issue to Scott as we've discussed his previous experience with Mermaid JS diagrams.

smckeon-so:
@ycarmy I don't have permissions to edit the Visio diagram. We need to change APS to AFC.

ycarmy:
> @ycarmy I don't have permissions to edit the Visio diagram. We need to change APS to AFC.

@gknopf-aranya I believe you are the owner of that diagram. Can you please give Scott permission to edit?

gknopf-aranya:
I provided a link with write permissions to @smckeon-so  in a DM. Didn't want to give the entire internet write access here.

PR #20, opened 2025-01-09
5/Add aranya product diagram

smckeon-so:
This PR adds the Aranya architecture diagram as a Mermaid JS code block. It also adds support for the `elk` layout engine which will make some of the more complex diagrams more readable. 

See https://aranya-project.github.io/aranya-docs/aranya-architecture/ for the deployed page. Currently the Mermaid rendering for this diagram isn't a quick as I'd like, however once cached it's not bad. We'll want to improve that in the future. 

Closes https://github.com/aranya-project/aranya-docs/issues/5

~~The downside to this is, Mermaid is limiting in the way it automatically flows the nodes. We can make some improvements by using the alternate layout engine `elk`. However that is not available in the default GH Markdown engine that GH uses to render this in the repo. We can however install elk in the published GH Pages docs using the `@mermaid-js/layout-elk` npm package.  But that wouldn't do anything for the vanilla rendering of the diagram in the repo, only the deployed Jekyll site.~~

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -36,8 +36,4 @@
             ga('send', 'pageview');
         </script>
     {% endif %}
-
-    {% if page.mermaid %}
-        {% include mermaid.html %}
-    {% endif %}
 </head>
--- a/_includes/mermaid.html
+++ b/_includes/mermaid.html
@@ -1,3 +1,0 @@
-<script type="text/javascript"
-  src="https://unpkg.com/mermaid@11.4.1/dist/mermaid.min.js">
-</script>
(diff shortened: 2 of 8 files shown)


nikki-aranya:
This probably won't be the only graph we want to add so it might be worth the effort to get elk set up now

nikki-aranya (approved):
From my primitive knowledge of JS, this seems like an excellent solution. You're right that elk is a way better engine for our current graphs. This renders as text, it would be nice to have the other diagrams as an svg so we can render them similarly but that's a different problem to solve.

jdygert-spok (approved):

PR #23, opened 2025-01-10
Add jekyll gh pages action

smckeon-so:
This PR will add the jekyll GH action to deploy to GH pages.

### DIFF
--- a/.github/workflows/jekyll-gh-pages.yml
+++ b/.github/workflows/jekyll-gh-pages.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Build with Jekyll
         uses: actions/jekyll-build-pages@v1
         with:
-          source: ./
+          source: ./docs
           destination: ./_site
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v3
--- a/.github/workflows/jekyll-gh-pages.yml
+++ b/.github/workflows/jekyll-gh-pages.yml
@@ -1,0 +1,51 @@
+# Sample workflow for building and deploying a Jekyll site to GitHub Pages
+name: Deploy Jekyll with GitHub Pages dependencies preinstalled
+
+on:
+  # Runs on pushes targeting the default branch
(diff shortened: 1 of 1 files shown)


gknopf-aranya (changes requested):
Let's add an `index.md`

jdygert-spok:
Isn't there a builtin jekyll action, so we don't need to add one?

gknopf-aranya:
> Isn't there a builtin jekyll action, so we don't need to add one?

In the settings, when you click to configure Jekyll it helps you setup a default workflow. Perhaps that's what they meant by built-in? I initially thought it would be hidden and just work. With settings managed via the GitHub settings page.
https://github.com/aranya-project/aranya-docs/settings/pages

jdygert-spok:
I think `deploy from a branch` uses the builtin one. (Just switched it, didn't think it'd save immediately....)

jdygert-spok:
Yeah, that's what it must have done when the previous PR was merged. E.g. it built https://aranya-project.github.io/aranya-docs/afc even though we didn't have an action setup.

jdygert-spok:
Here you can see the default action being ran: https://github.com/aranya-project/aranya-docs/actions/runs/12714260964/job/35443948351

gknopf-aranya:
Let's remove the workflow from this PR then and just add the `index.md` to it. Then changes the PR title and description to reflect that.

smckeon-so:
Yes, we shouldn't need this anymore. Closing to use the built in one. We may need this action though if/when we want to change the theme.

jdygert-spok:
I think most of everything you can just set in the config: https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll#configuring-jekyll-in-your-github-pages-site

gknopf-aranya:
Can we put our logo in the default-action without needing to setup our own workflow?

PR #29, opened 2025-01-16
AQC spec

benz-spideroak:


docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
+# Aranya Quic Channels
+
+## Overview
+
+Aranya Quic Channels are a way for application data to be sent over the 
+network to other peers using tls to ensure security. The QUIC protocol will
+be used to send data. This provides multiplexing of connections, tls for
+security, and custom congestion control.
```
gknopf-aranya, line 8: Some more things we could mention here:
Which congestion control algorithm we're selecting initially
A reference to some more info about the algorithm
Whether it's built into s2n_quic and a link to the docs on how to configure it (we should provide a link the the s2n_quic repo somewhere)

docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
...
+Quic channels will be stored in an `FnvIndexMap` with a max size of `32`. 
+Creating a channel when the map is full will fail with a `QuicChannelError`.
+
+```rust
+/// FNVIndexMap requires that the size be a power of 2.
+const MAXIMUM_CHANNELS: usize = 32;
+type ChannelMap = FNVIndexMap<(SocketAddr, GraphID), BiDirectionalStream, MAXIMUM_CHANNELS>
+```
+
+BiDirectional streams will have the following methods, mirroring the API of
```
gknopf-aranya, line 21:
```suggestion
Bidirectional streams will have the following methods, mirroring the API of
```

docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
...
+pub async fn receive(&mut self) -> Result<Option<bytes::Bytes>>
+pub async fn send(&mut self, mut data: bytes::Bytes) -> Result<()>
+```
+
+## API Methods
+
+Methods needed by the daemon's Unix domain socket API to support sending and receiving fast channel messages.
+
+- `CreateChannel(team_id: GraphID, addr: SocketAddr) -> Result<BiDirectionalStream, QuicChannelError>` creates a channel in Aranya, returning a `BiDirectionalStream`. Adds the stream to the `ChannelMap`.
+- `DeleteChannel(team_id: GraphID, addr: SocketAddr)` closes the channel if it exists and is open and removes the channel from the `ChannelMap`.
```
gknopf-aranya, line 35: There could be more than one channel created with a certain peer. So it would be better to accept the `ChannelID` here.

docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
...
+```rust
+pub async fn receive(&mut self) -> Result<Option<bytes::Bytes>>
+pub async fn send(&mut self, mut data: bytes::Bytes) -> Result<()>
+```
+
+## API Methods
+
+Methods needed by the daemon's Unix domain socket API to support sending and receiving fast channel messages.
+
+- `CreateChannel(team_id: GraphID, addr: SocketAddr) -> Result<BiDirectionalStream, QuicChannelError>` creates a channel in Aranya, returning a `BiDirectionalStream`. Adds the stream to the `ChannelMap`.
```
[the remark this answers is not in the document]
gknopf-aranya: Unless we're planning to multiplex multiple labels/channels over the same QUIC stream?

docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
...
+BiDirectional streams will have the following methods, mirroring the API of
+s2n_quic. Receive will return a chunk of data from the stream. It will 
+return Ok(None) when the stream is finished.
+
+```rust
+pub async fn receive(&mut self) -> Result<Option<bytes::Bytes>>
+pub async fn send(&mut self, mut data: bytes::Bytes) -> Result<()>
+```
+
+## API Methods
```
gknopf-aranya, line 30: Missing `ReceiveSessionCommand()` from the afc spec. Or are we assuming those API methods already exist and adding new ones for QUIC channels? If this is the case, we should document in this spec that it is an addition to the afc spec. We'd need to use different names for `CreateChannel()` and `DeleteChannel()` if that's the case since those already exist in the afc spec.

docs/quic-channels.md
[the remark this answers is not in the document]
elagergren-spideroak: Yes, I think we should have the label here too.

We'll (probably) eventually want to multiplex multiple channels over the same QUIC connection (not stream!), but that's an implementation detail.

docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
...
+type ChannelMap = FNVIndexMap<(SocketAddr, GraphID), BiDirectionalStream, MAXIMUM_CHANNELS>
+```
+
+BiDirectional streams will have the following methods, mirroring the API of
+s2n_quic. Receive will return a chunk of data from the stream. It will 
+return Ok(None) when the stream is finished.
+
+```rust
+pub async fn receive(&mut self) -> Result<Option<bytes::Bytes>>
+pub async fn send(&mut self, mut data: bytes::Bytes) -> Result<()>
```
elagergren-spideroak, line 27: `data` should be `&[u8]`

docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
...
+const MAXIMUM_CHANNELS: usize = 32;
+type ChannelMap = FNVIndexMap<(SocketAddr, GraphID), BiDirectionalStream, MAXIMUM_CHANNELS>
+```
+
+BiDirectional streams will have the following methods, mirroring the API of
+s2n_quic. Receive will return a chunk of data from the stream. It will 
+return Ok(None) when the stream is finished.
+
+```rust
+pub async fn receive(&mut self) -> Result<Option<bytes::Bytes>>
```
elagergren-spideroak, line 26: `receive` should write _into_ a buffer, like `recv(&mut self, buf: &mut [u8]) -> Result<usize, Error>`

docs/quic-channels.md
gknopf-aranya: Since I believe we're planning to keep AFC as an experimental feature, let's call the quic channel methods `CreateQuicChannel`, `DeleteQuicChannel`, etc.

docs/quic-channels.md
gknopf-aranya, line 26: Let's mention that the API methods in this section will be added to the user library's application API.

docs/quic-channels.md
jdygert-spok: s2n-quic is designed around `Bytes`. Even if you use `AsyncWrite` it just copies it to a `Bytes`: https://github.com/aws/s2n-quic/blob/adc7ba917ba5c58a0835ba53c6e17656aa81192b/quic/s2n-quic/src/stream/send.rs#L450

docs/quic-channels.md
```diff
@@ -0,0 +1,35 @@
+# Aranya Quic Channels
```
gknopf-aranya, line 1: Are we able to reuse `TxpMsg` from the AFC spec, or is there anything we'd like to change in the packets sent via QUIC channels? E.g. `AqcId` instead of `AfcId`.

gknopf-aranya, line 1: Let's mention that a new `shm` interface between the user library and daemon is being added to store the quic channel keys.

docs/quic-channels.md
gknopf-aranya: From my notes: "s2n_quic has a non-loss based congestion control algorithm"

docs/quic-channels.md
gknopf-aranya, line 1: @elagergren-spideroak should we add a section to this spec about the custom cryptography we're going to use with the QUIC channels? Or will that be in a different spec?

docs/quic-channels.md
elagergren-spideroak: Ugh, so each read/write incurs an allocation?

docs/quic-channels.md
benz-spideroak: More than one channel per team?

elagergren-spideroak: Yes. A channel is 1:1 between two peers, A and B. Both A and B can have multiple channels between each other. Each channel has a unique ID derived from the cryptographic secrets.

docs/quic-channels.md
benz-spideroak: What keys will be stored?

elagergren-spideroak: Cryptographic keys :)

gknopf-aranya, line 1: @elagergren-spideroak, @ycarmy will we need any policy updates to support quic channels?
|https://github.com/aranya-project/aranya/blob/main/crates/aranya-daemon/src/policy.md
It looks like we could mostly reuse the AFC policy, although we'd want to have a way to differentiate which shm path the keys are going to after processing the ephemeral effects:
https://github.com/aranya-project/aranya/blob/main/crates/aranya-daemon/src/api.rs#L222

We could create different action/effects for this or add a type parameter to specify afc/aqc.

docs/quic-channels.md
jdygert-spok: Yeah, and it's deeply baked in.

docs/quic-channels.md
benz-spideroak: Why is `ReceiveSessionCommand` needed?

docs/quic-channels.md
benz-spideroak: They won't be though. These are methods on the `BiDirectionalStream` struct.

docs/quic-channels.md
benz-spideroak: I don't think any of that is needed. QUIC will handle the setup and encryption of the connection.

benz-spideroak:
We had talked about a stream based API, but I'm not sure how we do that when there can be many channels per connection. To support streaming we'd need a connection for each channel.

docs/quic-channels.md
```diff
@@ -0,0 +1,86 @@
...
+
+Quic connections will be stored in an `FnvIndexMap` with a max size of `32`. 
+Creating a connection when the map is full will close and remove the 
+connection that has gone the longest without being used.
+A connection allows two peers to communicate and can be used for any number of 
+channels.
+
+```rust
+/// FNVIndexMap requires that the size be a power of 2.
+const MAXIMUM_CONNECTIONS: usize = 32;
```
benz-spideroak, line 21: Is it worth limited the number of connections when we can't avoid allocating to store the messages unless we're ok dropping old messages.

docs/quic-channels.md
```diff
@@ -0,0 +1,86 @@
+# Aranya Quic Channels
```
smckeon-so, line 1: To show up in the published doc site menu, this will need a Front Matter block as the first line.
```
---
layout: page
title: Aranya Quic Channels
permalink: "/aranya-quic-channels/"
---
```

docs/quic-channels.md
elagergren-spideroak: this spec

elagergren-spideroak: yes, we'll need policy updates

docs/quic-channels.md
elagergren-spideroak: We shouldn't need to store any messages, I don't think. When the user calls `aqc_poll` (*) we check the (next) connection to see if there is any data. If so, we report that to the user. Then the user calls `aqc_recv` (*) and we receive the data from the connection and hand it to the user.

Basically, we're a very thin wrapper around QUIC.

(*): We can use different names, of course.

benz-spideroak: I'm not understanding the usefulness of separate poll and receive calls with QUIC. If there's data to receive why not return it immediately? 

If we store the messages it allows a nicer API for the user. They can send to a channel as often as they want. And the messages are available to read on the other side whenever they're ready for them.  The code to receive messages can be async and handle communication transparently to the user.

Without that, you have to block on sending until the other side calls receive and block on receiving until a message arrives.

elagergren-spideroak: > I'm not understanding the usefulness of separate poll and receive calls with QUIC. If there's data to receive why not return it immediately?

I'm fine with `aqc_recv` immediately returning data.

> If we store the messages it allows a nicer API for the user. They can send to a channel as often as they want. And the messages are available to read on the other side whenever they're ready for them. The code to receive messages can be async and handle communication transparently to the user.

I guess I'm not understanding how the API is storing messages. Are you proposing that we have a background thread that reads/writes to the connections, inserts the messages into a queue, and `aqc_recv` just reads from that queue?

benz-spideroak: > I guess I'm not understanding how the API is storing messages. Are you proposing that we have a background thread that reads/writes to the connections, inserts the messages into a queue, and aqc_recv just reads from that queue?

Essentially, yes. With QUIC you first open a connection, the server awaits the client opening a stream, then the server and client send and receive from the stream. It's async so the server is always waiting for new connections and for the client to open a stream using the connection. It seems odd to me to not read from the stream when you know it's available.  

It seems hard to get around some amount of buffering messages. The server code will be running in a spawned async thread so it can't just write the message directly to a buffer. It'll have to copy it to some shared memory that `receive` can read from. We could only buffer one message at a time. That would remove the need to allocate, but it would add network latency. It would also make it impossible to quickly send a bunch of messages and to receive messages from multiple peers at once.

docs/quic-channels.md
gknopf-aranya: Do you mean this will be handled in the crypto portion of this spec?

docs/quic-channels.md
gknopf-aranya: `ReceiveSessionCommand()` is where ephemeral commands for creating AFC or QUIC channels are passed from the user library to the daemon. The user library received the command via the network from another peer, then passes it into the daemon so it can be added onto an ephemeral branch of the Aranya graph.

docs/quic-channels.md
benz-spideroak: I don't think we'll use this at all. `TxpMsg::Data` has been replaced by `QuicMessage::ChannelMessage`. Once the QUIC connection is set up we won't need `TxpMsg::Ctrl` and we can't send `TxpMsg::Ctrl` before the connection is set up. We will need to share the crypto keys somehow, but it'll be done differently than with AFC.

docs/quic-channels.md
benz-spideroak: That may be needed, but it'll be a part of the crypto portion of the spec.

docs/quic-channels.md
gknopf-aranya: @elagergren-spideroak, @ycarmy is there a spec where we want mention those policy changes?

gknopf-aranya: We'll want to document that `AqcCtrl` and `AqcId` are passed around via the QUIC transport during setup of the AQC channel.

gknopf-aranya: `AqcCtrl` would be the ephemeral command that is added to both peer's Aranya sessions in order to setup an encrypted channel. Because it is off-graph, it cannot be passed via the Aranya sync protocol. So it needs to be passed around some other way. In the case of AFC, we used the same TCP protocol to pass the ephemeral command to initialize AFC channels as well as to send encrypted data afterwards. I think we'll want to do the same thing for QUIC channels. I don't see a reason to introduce a different transport mechanism for passing the ephemeral command around.

The `AqcId` is a unique identifier for the QUIC channel that can be used to lookup the crypto keys for open/seal (decrypt/encrypt) operations on the data. It will be sent along with the data via the QUIC channel, so the receiving peer can lookup the crypto keys to decrypt the data with.

gknopf-aranya: Here is some more info on Aranya sessions:
https://github.com/aranya-project/aranya-docs/blob/main/docs/aranya-sessions-note.md

gknopf-aranya: If it's possible to configure the crypto as part of a QUIC stream, I believe this means the ephemeral commands would be passed around ahead of time via a different QUIC stream, then crypto would be initialized and used to initialize a new QUIC stream with the encryption/decryption already set up to use the right keys for the AQC channel.

@elagergren-spideroak, @ycarmy do you have more thoughts on this?

benz-spideroak: > The `AqcId` is a unique identifier for the QUIC channel that can be used to lookup the crypto keys for open/seal (decrypt/encrypt) operations on the data. It will be sent along with the data via the QUIC channel, so the receiving peer can lookup the crypto keys to decrypt the data with.

We won't be encrypting or decrypting data with QUIC channels. QUIC will handle that for us.

benz-spideroak: > `AqcCtrl` would be the ephemeral command that is added to both peer's Aranya sessions in order to setup an encrypted channel.

It sounds like we may need `AqcCtrl` depending on the crypto design. In that case we'd need another set of keys to use to share the channel keys. Maybe the same keys we'll use for sync.

djohnson-spideroak:
Taking a look, might have to resume tomorrow

docs/quic-channels.md
ycarmy, line 1: Looks like there's a decent amount of detail still missing. I suppose most of it is supposed to come under the crypto section? Do we have an ETA for when that might get added? (cc: @elagergren-spideroak)

ycarmy (commented):
Is there / will there be a separate spec for doing quic-based syncing of the graph?

docs/quic-channels.md
```diff
@@ -0,0 +1,92 @@
+---
+layout: page
+title: Aranya Quic Channels
+permalink: "/aranya-quic-channels/"
+---
+
+# Aranya Quic Channels
+
+## Overview
```
gknopf-aranya, line 9: @ycarmy, @benz-spideroak, @elagergren-spideroak, I was just discussing with @jkim-so how in the past we have occasionally pinned QUIC client connection ports to the same port as the QUIC server. This made it simpler to configure firewall and routing rules when deploying to a network. We may want to have the option of configuring the client to bind to any available port, pin it to the same port as the server, or pin it to a different port from the server.

gknopf-aranya:
> Is there / will there be a separate spec for doing quic-based syncing of the graph?

@ycarmy that's my understanding. This spec is for the QUIC transport equivalent of AFC. QUIC based Aranya graph syncing would be a different spec/implementation effort and is post-MVP.

docs/quic-channels.md
```diff
@@ -0,0 +1,81 @@
...
+/// Identifies a unique channel between two peers.
+pub struct QuicChannel {
+    /// The node id of the peer.
+    node_id: NodeId,
+    /// The channel label. This allows multiple channels between two peers.
+    label: Label,
+}
+
+/// FNVIndexMap requires that the size be a power of 2.
+const MAXIMUM_CONNECTIONS: usize = 32;
```
gknopf-aranya, line 46: May want to start with a number larger than this. Such as 128.

docs/quic-channels.md
gknopf-aranya, line 1: I don't see any mention of IP addresses or network identifiers (from the MVP spec) in this spec. I would expect there to be a mention somewhere of what mechanism is being used to bind QUIC servers to ports and multiplex incoming peer connections based on their network address.

benz-spideroak: We're not using network address to multiplex connections. We're using the TLS key they connect with. Each channel will have a unique key.

docs/quic-channels.md
jdygert-spok: https://groups.google.com/a/chromium.org/g/proto-quic/c/Zz6f-F3LJRs and https://quic-go.net/docs/quic/transport/ say it is possible, so as long as s2n-quic doesn't do something weird, it should work

docs/quic-channels.md
```diff
@@ -0,0 +1,81 @@
...
+/// from the ConnectionMap. It will then send the message to the peer.
+SendQuicMessage(channel: QuicChannel, message: &[u8])
+    -> Result<(), QuicMessageError>
+/// Returns the next message from qc_receiver or none if the channel is empty. 
+ReceiveQuicData(target: &mut [u8])
+    -> Result<Option<(QuicChannel, usize)>, QuicMessageError>  
+/// Closes the connection if it exists and is open and removes the connection 
+/// from the `ConnectionMap`.
+CloseQuicConnection(addr: SocketAddr) 
+```
```
ycarmy, line 81: So will there be no polling?

elagergren-spideroak: yeah, we should use AQC, Aqc, etc. instead of QUIC, Quic, etc.

docs/quic-channels.md
```diff
@@ -0,0 +1,81 @@
...
+receiving Quic channel messages.
+
+```rust
+/// This will either create a new connection or retrieve an existing connection
+/// from the ConnectionMap. It will then send the message to the peer.
+SendQuicMessage(channel: QuicChannel, message: &[u8])
+    -> Result<(), QuicMessageError>
+/// Returns the next message from qc_receiver or none if the channel is empty. 
+ReceiveQuicData(target: &mut [u8])
+    -> Result<Option<(QuicChannel, usize)>, QuicMessageError>
```
ycarmy, line 77: Is there a particular reason the first API has "Message" while the second uses "Data"?

gknopf-aranya:
@djohnson-spideroak are we reusing AFC labels for ACQ or using different labels?

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+title: Aranya Quic Channels
+permalink: "/aranya-quic-channels/"
+---
+
+# Aranya Quic Channels (AQC)
+
+## Overview
+
+Aranya Quic Channels are a way for application data to be sent over the 
+network to other peers using TLS to ensure security. The QUIC protocol will
```
gknopf-aranya, line 12: Mentioning TLS without mentioning that we are using custom crypto is a bit misleading.
A reader could assume that certificates will be used with the default TLS cryptography.
We should reference the crypto section of this spec here.

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+    pub new_channels: HVec<AqcChannel, MAXIMUM_UNRECEIVED_CHANNELS>,
+}
+
+impl AqcClient {
+    /// Create an Aqc client with the given certificate chain.
+    pub fn new<T: provider::tls::Provider>(cert: T) -> Result<AqcClient, AqcError> 
+
+    /// Receive the next available channel. If no channel is available, return None.
+    /// This method will return a channel created by a peer that hasn't been received yet.
+    pub fn receive_channel(&mut self) -> Option<AqcChannel>
```
gknopf-aranya, line 40: This should only receive channels that are permitted by the policy.

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+impl AqcClient {
+    /// Create an Aqc client with the given certificate chain.
+    pub fn new<T: provider::tls::Provider>(cert: T) -> Result<AqcClient, AqcError> 
+
+    /// Receive the next available channel. If no channel is available, return None.
+    /// This method will return a channel created by a peer that hasn't been received yet.
+    pub fn receive_channel(&mut self) -> Option<AqcChannel> 
+
+    /// Create a new channel to the given address.
+    pub async fn create_channel(&mut self, addr: SocketAddr, label: Label) -> Result<AqcChannel, AqcError>
```
gknopf-aranya, line 43: This should only create channels that are permitted by the policy.

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+
+## Channels
+
+All peers will spawn an async task to await new connections.
+
+A connection allows two peers to communicate and will be used for one channel.
+When receiving a QUIC connection the TLS key is used to connect will identify 
+the channel.
```
gknopf-aranya, line 53:
```suggestion
When receiving a QUIC connection, the TLS key used to connect will identify 
the channel.
```

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+
+When a channel is created the peer who creates the channel will connect to
+the other peer and keep the connection alive.
```
gknopf-aranya, line 56: How is the connection kept alive? Is data periodically sent to ensure the connection doesn't time out?

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+
+    /// Creates a new unidirectional stream for the channel. 
+    pub async fn create_unidirectional_stream(&mut self) -> Result<AqcSendStream, AqcError> 
+
+    /// Creates a new bidirectional stream for the channel. 
+    pub async fn create_bidirectional_stream(
+        &mut self,
+    ) -> Result<(AqcSendStream, AqcReceiveStream), AqcError> {
+
+    /// Close the channel if it's open. If the channel is already closed, do nothing.
```
gknopf-aranya, line 102: Could mention here that it closes the corresponding QUIC connection

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+Aranya Quic Channels are a way for application data to be sent over the 
+network to other peers using TLS to ensure security. The QUIC protocol will
+be used to send data. This provides multiplexing of connections, TLS for
+security, and custom congestion control. We will default to using BBRv2
+for congestion control.
+
+## Clients
+
+Each peer will have an AqcClient. This can be used to create new channels
+or receive incoming channels. A channel represents a quic connection
```
gknopf-aranya, line 20: @benz-spideroak, @elagergren-spideroak will the QUIC connection be established at the same time the AQC channel is created?
Or should we lazily open the QUIC connection when the initial data is sent?

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+    /// Returns the new channel
+    pub fn new(conn: Handle) -> Self
+
+    /// Returns a bidirectional stream if one has been received. 
+    /// If no stream has been received return None. 
+    pub async fn receive_bidirectional_stream(&mut self) -> Option<(AqcSendStream, AqcReceiveStream)> 
+
+    /// Returns a unidirectional stream if one has been received. 
+    /// If no stream has been received return None. 
+    pub async fn receive_unidirectional_stream(
```
gknopf-aranya, line 90: Should mention that the stream will not be received if the direction is incorrect according to the policy

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+## Streams
+A stream allows sending or receiving data. 
+
+QUIC will split messages into several pieces so a data chunk in a stream 
+may not represent a complete message.
+
+```rust
+pub struct AqcReceiveStream {
```
gknopf-aranya, line 114: Which object will maintain all the open streams?
I couldn't find this in the spec.

gknopf-aranya (dismissed):
lgtm pending resolving the open feedback

docs/quic-channels.md
djohnson-spideroak: Yeah, should mention we are using aranya crypto with TLS

docs/quic-channels.md
benz-spideroak: It will return an error if the channel isn't allowed by policy.

docs/quic-channels.md
benz-spideroak: You call keep_alive on the connection and the library handles that.

docs/quic-channels.md
gknopf-aranya: The crypto section has more detail. But we should say something like this in the overview:
"using our custom TLS crypto"

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
+---
+layout: page
+title: Aranya Quic Channels
+permalink: "/aranya-quic-channels/"
+---
+
+# Aranya Quic Channels (AQC)
```
gknopf-aranya, line 7:
```suggestion
# Aranya QUIC Channels (AQC)
```

docs/quic-channels.md
benz-spideroak: We can't connect until the peer has the PSK. Once the key is shared we'll open the channel and keep it open. In the initial version I'll do whatever is simplest and we can optimize later.

docs/quic-channels.md
gknopf-aranya: I mean in the docs I think we should mention that the operation will only succeed if it is permitted by the policy.

docs/quic-channels.md
gknopf-aranya: Ideally with a link to the crypto section.

docs/quic-channels.md
gknopf-aranya, line 1: @elagergren-spideroak, @benz-spideroak, @djohnson-spideroak, @mooreoak we've made some design decisions regarding AQC PSK key management in the Aranya client and daemon that I'd like to capture in this spec:
- Passing PSK encap from daemon -> client via the TARPC IPC API
- Passing PSK secret from daemon -> client via the key store

elagergren-spideroak: Maybe. I'm sorta iffy about including it here since it's something we very likely might change once we update the IPC.

elagergren-spideroak: It's also kinda/sorta an implementation detail?

gknopf-aranya: Where else would we document this? It is AQC related and has security implications.
Agree that the specific IPC mechanisms could change later.

gknopf-aranya: More generally, the secret should only be passed via a secured/encrypted API. The encap can be passed over an unsecured API.

elagergren-spideroak: > Where else would we document this? It is AQC related and has security implications.

In the daemon's documentation. The AQC spec doesn't rely on having a daemon. You should be able to take the spec and implement AQC on top of an ad-hoc usage of `aranya-core`, for example. We shouldn't have to update the AQC spec if we decide to change how the daemon works.

> More generally, the secret should only be passed via a secured/encrypted API. The encap can be passed over an unsecured API.

I can document this, yeah. And to some extent I believe we already do—we mention that the encapsulated KEM shared secret can be publicly shared; indeed, it is sent over the graph.

gknopf-aranya: I'll add some more docs about this to the client/daemon code.

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+```policy
+use aqc
+use crypto
+use device
+use envelope
+use idam
+use perspective
+
+// A device.
+struct Device {
```
gknopf-aranya, line 596: The current policy uses the `Device` fact struct rather than defining another struct.
```
// A device on the team.
fact Device[device_id id]=>{role enum Role, sign_key_id id, enc_key_id id}
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+    // The device ID of the participant that can read data.
+    reader_id id,
+    // The channel author's encryption key ID.
+    author_enc_key_id id,
+    // The channel peer's encoded public encryption key.
+    peer_enc_pk bytes,
+    // The channel label.
+    label int,
+    // The channel's unique identifier.
+    channel_id,
```
gknopf-aranya, line 927: Missing type
```suggestion
    channel_id id,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+command AqcCreateBidiChannel {
+    fields {
+        // The channel peer's device Id.
+        peer_id id,
+        // The label applied to the channel.
+        label int,
+        // The channel peer's encapsulated KEM shared secret.
+        peer_encap bytes,
+        // The channel's unique identifier.
+        channel_idid,
```
gknopf-aranya, line 811:
```suggestion
        channel_id id,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+    // The channel author's encryption key ID.
+    author_enc_key_id id,
+    // The channel peer's device Id.
+    peer_id id,
+    // The channel peer's encoded public encryption key.
+    peer_enc_pk bytes,
+    // The channel label.
+    label int,
+    // The channel's unique identifier.
+    channel_id,
```
gknopf-aranya, line 779:
```suggestion
    channel_id id,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+// Returned by `create_bidi_channel`.
+struct AqcBidiChannel {
+    // The channel peer's encapsulated KEM shared secret.
+    peer_encap bytes,
+    // The channel's unique ID.
+    channel_id id,
+}
+
+// Creates a bidirectional AQC channel.
+function create_bidi_channel(
```
gknopf-aranya, line 1057: I copied this into `policy.md` and it's expecting an implementation for this function.
Is this AQC FFI code just here for reference and not intended to be used directly as policy?

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+
+#### AQC FFI
+
+```policy
+// Returned by `create_bidi_channel`.
+struct AqcBidiChannel {
+    // The channel peer's encapsulated KEM shared secret.
+    peer_encap bytes,
+    // The channel's unique ID.
+    channel_id id,
```
gknopf-aranya, line 1053: Confirming this is similar to what used to be the `channel_key_id` in AFC?

gknopf-aranya: It's a little confusing to call this a `channel_id` in some places and a `key_id` in others. E.g. `aranya_aqc_util::BidiChannelCreated`:
```
    /// The unique key identifier for the
    /// [`BidiAuthorSecret`][aranya_crypto::aqc::BidiAuthorSecret].
    pub key_id: BidiKeyId,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+    policy {
+        let author = get_valid_device(envelope::author_id(envelope))
+        let label_id = envelope::command_id(envelope)
+
+        // Verify that the label does not already exist.
+        //
+        // This will happen in the `finish` block if we try to
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        check !exists AqcLabel[name: this.name, author_id: author.devce_id]
```
gknopf-aranya, line 1117:
```suggestion
        check !exists AqcLabel[name: this.name, author_id: author.device_id]
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+
+        finish {
+            create AqcLabel[name: this.name, author_id: author.devce_id]=>{label_id: label_id}
```
gknopf-aranya, line 1120:
```suggestion
            create AqcLabel[name: this.name, author_id: author.device_id]=>{label_id: label_id}
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+        let author = get_valid_device(envelope::author_id(envelope))
+
+        // NB: Check roles, other ACLs here.
+
+        // Verify that the label exists.
+        //
+        // This will happen in the `finish` block if we try to
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        let label = check_unwrap query AqcLabel[name: this.name, author_id: label_author.devce_id]
```
gknopf-aranya, line 1172:
```suggestion
        let label = check_unwrap query AqcLabel[name: this.name, author_id: label_author.device_id]
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+
+        finish {
+            delete AqcLabel[name: this.name, author_id: label_author.devce_id]=>{}
```
gknopf-aranya, line 1175:
```suggestion
            delete AqcLabel[name: this.name, author_id: label_author.device_id]=>{}
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        check !exists AqcLabel[name: this.name, author_id: author.devce_id]
+
+        finish {
+            create AqcLabel[name: this.name, author_id: author.devce_id]=>{label_id: label_id}
+
+            emit LabelCreated {
+                name: this.name,
+                author_id: author.device_id,
```
gknopf-aranya, line 1124:
```suggestion
                label_author_id: author.device_id,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+) struct AqcUniChannel
+```
+
+### Labels
+
+```policy
+// Records an AQC label.
+//
+// `author_id` is the ID of the device that created the label.
+fact AqcLabel[name string, author_id id]=>{label_id id}
```
gknopf-aranya, line 1091:
```suggestion
fact AqcLabel[name string, label_author_id id]=>{label_id id}
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+        //
+        // This will happen in the `finish` block if we try to
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        let label = check_unwrap query AqcLabel[name: this.name, author_id: label_author.devce_id]
+
+        finish {
+            delete AqcLabel[name: this.name, author_id: label_author.devce_id]=>{}
+
+            emit AqcLabelCreated {
```
gknopf-aranya, line 1177: effect has 3 fields:
```
effect AqcLabelCreated {
    // The label name.
    name string,
    // The ID of the device that created the label.
    label_author_id id,
    // Uniquely identifies the label.
    label_id id,
}
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+    }
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        finish {
+            emit QueriedAqcLabel {
+                name: this.name,
+                label_author_id: this.author_id,
```
gknopf-aranya, line 1225:
```suggestion
                label_author_id: this.label_author_id,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+        peer_enc_pk,
+        peer_id,
+        label,
+    )
+
+    publish AqcCreateBidiChannel {
+        peer_id: peer_id,
+        label: label,
+        peer_encap: ch.peer_encap,
+        channel_id: ch.channel_id,
```
gknopf-aranya, line 758: FFI is currently using `key_id` field:
```suggestion
        channel_id: ch.key_id,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+        reader_id,
+        label,
+    )
+
+    publish AqcCreateUniChannel {
+        writer_id: writer_id,
+        reader_id: reader_id,
+        label: label,
+        peer_encap: ch.peer_encap,
+        channel_id: ch.channel_id,
```
gknopf-aranya, line 904: FFI on `issue/183` branch is currently using `key_id` field:
```suggestion
        channel_id: ch.key_id,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+
+```policy
+// Records an AQC label.
+//
+// `author_id` is the ID of the device that created the label.
+fact AqcLabel[name string, author_id id]=>{label_id id}
+
+action create_aqc_label(name string) {
+    publish CreateAqcLabel {
+        name: label,
```
gknopf-aranya, line 1095:
```suggestion
        name: name,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+    label_id id,
+    // The ID of the device that deleted the label.
+    author_id id,
+}
+
+// Emits `QueriedAfcLabel` for each label with the same name.
+action query_aqc_label(name string) {
+    map AqcLabel[name: name, author_id: ?] as f {
+        publish QueryAqcLabel {
+            label: f.name,
```
gknopf-aranya, line 1204:
```suggestion
            name: f.name,
```

gknopf-aranya:
I've copied the AQC policy from this spec to my AQC policy PR and made a few changes to get it to compile:
https://github.com/aranya-project/aranya/pull/145/commits/b2a0e44eb6aa705300ab0cd78c1f27addc6f2184

docs/quic-channels.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Clients
+
+Each peer will have an AqcClient. This can be used to create new channels
+or receive incoming channels. A channel represents a quic connection
+between two peers that is secured by crypto generated for the channel.
+
+```rust
+/// The maximum number of channels that haven't been received.
+const MAXIMUM_UNRECEIVED_CHANNELS: usize = 20;
```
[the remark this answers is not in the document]
benz-spideroak: No. This is the number of channels this client has received, but client.receive_channel hasn't been called for them yet.

docs/quic-channels.md
elagergren-spideroak: I know, I just stubbed out all the external bits.

docs/quic-channels.md
elagergren-spideroak: FFI definitions are not included in policy, no. So, please do not copy the FFI section into `policy.md`.

docs/quic-channels.md
elagergren-spideroak: Yes, iThe term "key id" came from previous discussions Moore and I had about AFC where each packet would identity the current key being used.

AQC doesn't do that, so it will use "channel id" instead. I will update `aranya-aqc-util` accordingly.

docs/quic-channels.md
```diff
@@ -0,0 +1,1459 @@
...
+    parent_cmd_id id,
+    author_enc_key_id id,
+    their_pk bytes,
+    seal_id id,
+    open_id id,
+    label int,
+) struct AqcUniChannel
+```
+
+### Labels
```
gknopf-aranya, line 1085: Should there be an `AqcAssignLabel` command in this section?

gknopf-aranya: AFC had an `AssignLabel` label command and the label assignments were checking during channel creation.

elagergren-spideroak: The policy listed here _does_ check label assignments.

I will add the "assign label" bits.

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+These are consistent with the recommendations for non-key pair
+authentication in [[RFC 9180]] section 5.1.3 and [[AKE]].
+
+The following contextual binding is used when deriving the PSK
+from the HPKE encryption context:
+
+- The strings "aqc bidi psk" and "aqc uni psk" bind the PSK to
+  either bidirectional or unidirectional channels. This further
+  ensures that a bidirectional channel cannot be substituted for
+  a unidirectional channel and vice versa.
```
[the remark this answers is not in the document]
elagergren-spideroak: What do you mean?

docs/quic-channels.md
```diff
@@ -0,0 +1,1446 @@
...
+A, B, and C or D, E, and F will have their work disrupted.
+
+The approach outlined below side steps this problem, at the
+expense of a more verbose user experience.
+
+##### Design
+
+Labels are defined as a unique (name, author ID) tuple.
+
+TODO: expand more
```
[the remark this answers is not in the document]
elagergren-spideroak: Yes, I know, I'm currently writing this.

docs/quic-channels.md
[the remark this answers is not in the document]
djohnson-spideroak: from the start of this section:

```
##### Contextual Binding

The following contextual binding is used when creating the HPKE
encryption context:

- The strings "AqcBidiPsk" and "AqcUniPsk" bind the HPKE
  encryption context to a particular channel direction. This
  ensures that a bidirectional channel cannot be substituted for
  a unidirectional channel and vice versa.
```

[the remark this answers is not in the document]
elagergren-spideroak: Nope, the spec is correct.

"AqcBidiPsk" and "AqcUniPsk" are used when _creating_ the HPKE encryption context.

"aqc bidi psk" and "aqc uni psk" are used when _deriving the PSK from_ the HPKE encryption context.

docs/quic-channels.md
```diff
@@ -0,0 +1,1639 @@
...
+    seal_id id,
+    open_id id,
+    label_id id,
+) struct AqcUniChannel
+```
+
+### Labels
+
+```policy
+// Records a label for AQC and AFC.
```
gknopf-aranya, line 1082: @elagergren-spideroak, @djohnson-spideroak I thought we were making these label changes only for AQC. If we're also making them for AFC, then this creates more work related to updating AFC to use the new labels. Want to make sure we understand the implication and can plan to get that work done for MVP.

elagergren-spideroak: Labels are shared by both AQC and AFC. The security issues we identified with the old style of labels also applies to AFC.

docs/quic-channels.md
```diff
@@ -0,0 +1,1635 @@
...
+// Returns the channel operations that this device is allowed to
+// perform for a particular label.
+function get_allowed_op(device_id id, label_id id) enum ChanOp {
+    let assigned = check_unwrap query AssignedLabel[device_id: device_id, label_id: label_id]
+    return assigned.op
+}
+
+// Reports whether the devices have permission to create
+// a bidirectional AQC channel with each other.
+function can_create_aqc_bidi_channel(device1 id, device2 id, label_id id) bool {
```
gknopf-aranya, line 663: We used to check if the label was valid here when it was an int. That check doesn't make sense to do here anymore since we're using Aranya IDs. Although, it could be useful to check if the label exists before checking if devices have permission to use the label.

elagergren-spideroak: > We used to check if the label was valid here when it was an int. That check doesn't make sense to do here anymore since we're using Aranya IDs.

That's why I removed it :)

> Although, it could be useful to check if the label exists before checking if devices have permission to use the label.

I did not add `query Label[...]` to that function because I we already query the label _before_ calling that function, and querying twice is just needless overhead. I suppose we could make `can_create_aqc_bidi_label` lookup and return the label, but I'm not exactly sure how much I like that idea. Or we could make `can_create_aqc_bidi_label` take in a `label struct Label` instead of a `label_id id`, but I'm not really sure what that gets us.

Keep in mind that these APIs are a template for what will actually exist inside Aranya. For example, we have to check roles and whatnot. So, little things can be tweaked there.

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+        // results in a nicer error (I think?).
+        let label = check_unwrap query Label[label_id: this.label_id]
+
+        finish {
+            delete Label[label_id: label.label_id]=>{}
+
+            // Cascade the label assignments.
+            delete AssignedLabel[device_id: ?, label_id: label.label_id]
+
+            emit LabelCreated {
```
gknopf-aranya, line 1174: This is inside the `DeleteLabel` command:
```suggestion
            emit LabelDeleted {
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+    // Devices cannot create channels with themselves.
+    //
+    // This should have been caught by the AQC FFI, so check
+    // instead of just returning false.
+    check writer_id != reader_id
+
+    // The writer must have permissions to write (send) data.
+    let writer_op = get_allowed_op(writer_id, label_id)
+    match writer_op {
+        ChanOpt::ReadOnly => return false,
```
gknopf-aranya, line 697:
```suggestion
        ChanOp::ReadOnly => return false,
```
```
enum ChanOp {
    // The device can only receive data in channels with this
    // label.
    ReadOnly,
    // The device can only send data in channels with this
    // label.
    WriteOnly,
    // The device can send and receive data in channels with this
    // label.
    ReadWrite,
}
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+    // Devices cannot create channels with themselves.
+    //
+    // This should have been caught by the AQC FFI, so check
+    // instead of just returning false.
+    check writer_id != reader_id
+
+    // The writer must have permissions to write (send) data.
+    let writer_op = get_allowed_op(writer_id, label_id)
+    match writer_op {
+        ChanOp::ReadOnly => return false,
```
gknopf-aranya, line 697: I had to change to this to get it to compile:
```suggestion
ChanOp::ReadOnly => {return false},
```

```
error: failed to run custom build command for `aranya-daemon v0.5.1 (/Users/gknopf/scm/aranya/crates/aranya-daemon)`

Caused by:
  process didn't exit successfully: `/Users/gknopf/scm/aranya/target/debug/build/aranya-daemon-6d137b1eda75c5f4/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=src/policy.md

  --- stderr
  thread 'main' panicked at crates/aranya-daemon/build.rs:8:14:
  expected policy-ifgen to generate policy.rs: Syntax error: line 375 column 26:    --> 375:26
      |
  375 |         ChanOp::ReadOnly => return false,
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+    // Devices cannot create channels with themselves.
+    //
+    // This should have been caught by the AQC FFI, so check
+    // instead of just returning false.
+    check writer_id != reader_id
+
+    // The writer must have permissions to write (send) data.
+    let writer_op = get_allowed_op(writer_id, label_id)
+    match writer_op {
+        ChanOp::ReadOnly => { return false },
```
gknopf-aranya, line 697: Meant to suggest this earlier:
```
  thread 'main' panicked at crates/aranya-daemon/build.rs:8:14:
  expected policy-ifgen to generate policy.rs: Syntax error: line 375 column 45:    --> 375:45
      |
  375 |         ChanOp::ReadOnly => { return false },

```
```suggestion
        ChanOp::ReadOnly => { return false }
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+    label_name string,
+    // The ID of the author of the label.
+    label_author_id id,
+    // The ID of the device that revoked the label.
+    author_id id,
+}
+
+// Emits `QueriedLabelAssignment` for all labels the device has
+// been granted permission to use.
+action query_label_assignments(device_id) {
```
gknopf-aranya, line 1379:
```suggestion
action query_label_assignments(device_id id) {
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        let author = get_valid_device(envelope::author_id(envelope))
+
+        // Ensure that the author is one of the channel
+        // participants.
+        check author.device_id == this.writer_id ||
+              author.device_id == this.reader_id
```
gknopf-aranya, line 966: I'm looking at the ffi PR and it looks like `reader_id` was changed to `open_id`. `writer_id` was changed to `seal_id`.

gknopf-aranya: I must've thought the compiler was complaining about the field name changing when it was really a syntax error:
```
        let peer_id = if author.device_id == this.seal_id {
            this.open_id
        } else {
            this.seal_id
        }
``
```
  expected policy-ifgen to generate policy.rs: Syntax error: line 1972 column 13:     --> 1972:13
       |
  1972 |             this.open_id
       |             ^---
       |
       = expected struct_literal_field, function_call, action_call, publish_statement, let_statement, check_statement, match_statement, if_statement, finish_statement, map_statement, create_statement, update_statement, delete_statement, emit_statement, return_statement, or debug_assert
```

File: docs/quic-channels.md [hunk 55]
```diff
@@ -0,0 +1,1638 @@
...
+    // The channel label.
+    label_id id,
+    // The channel's unique identifier.
+    channel_id id,
+}
+
+// The effect that is emitted when the peer of a unidirectional
+// AQC channel successfully processes the `AqcCreateUniChannel`
+// command.
+effect AqcUniChannelReceived {
```
gknopf-aranya, line 924: I think the channel received effects need a `channel_id` field.
The `channel_id` field needs to be set on the `BidiChannelReceived` object when invoking the `handler.bidi_channel_received()` method.
Same for unidirectional channels.

gknopf-aranya: Or alternatively, the handler method shouldn't require the `channel_id` field to be set.

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+        let label_id = envelope::command_id(envelope)
+
+        // NB: Check roles, other ACLs here.
+
+        // Verify that the label does not already exist.
+        //
+        // This will happen in the `finish` block if we try to
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        check !exists Label[name: this.label_name, author_id: author.device_id]
```
gknopf-aranya, line 1116: Fact would need to be searched by the `label_id` key. But this check shouldn't be required since the command ID should always be unique each time `CreateLabel` is called.

docs/quic-channels.md
```diff
@@ -0,0 +1,1638 @@
...
+    // The ID of the author of the label.
+    label_author_id id,
+    // The ID of the device that revoked the label.
+    author_id id,
+}
+
+// Emits `QueriedLabelAssignment` for all labels the device has
+// been granted permission to use.
+action query_label_assignments(device_id id) {
+    map AssignedLabel[device_id: id, label_id: ?] as f {
```
gknopf-aranya, line 1380:
```suggestion
    map AssignedLabel[device_id: device_id, label_id: ?] as f {
```

docs/quic-channels.md
benz-spideroak: The stream will never be received if it doesn't pass the policy.

docs/quic-channels.md
elagergren-spideroak: The `*ChannelReceived` effects actually don't need the channel ID b/c it's derived from the encapsulation. I'll fix this up.

docs/quic-channels.md
elagergren-spideroak: I'm not sure what you mean. I just forgot to fix this line after the call last week.

docs/quic-channels.md
elagergren-spideroak: They've been updated.

docs/quic-channels.md
gknopf-aranya: I mean there doesn't seem to be a point to checking if the label ID fact already exists before creating it, when the ID is randomly generated and there is a very low probability of collisions.

docs/quic-channels.md
gknopf-aranya: How would the `channel_id` be derived from the encap?

docs/quic-channels.md
elagergren-spideroak: See my code comment that you highlighted. This sort of pattern is used throughout Aranya's policy and I don't want to deviate from it until I understand why it's being used.

Also: while the collision risk is obviously cryptographically negligible, perhaps the policy has a bug. E.g., maybe we use the wrong variable, like the author's ID or something.

docs/quic-channels.md
elagergren-spideroak: See the docs for  `BidiPeerEncap` and `UniPeerEncap`.

gknopf-aranya: `encap.id()` looks like the ID of the encap is the channel ID now:
```
    /// Uniquely identifies the bidirectional channel.
    #[inline]
    pub fn id(&self) -> BidiChannelId {
        BidiChannelId(Id::new::<CS>(self.as_bytes(), b"BidiChannelId"))
    }
```

elagergren-spideroak: It's always been like that. Sorry, isn't that what you asked for?

docs/quic-channels.md
```diff
@@ -0,0 +1,1687 @@
...
+    }
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        let author = get_valid_device(envelope::author_id(envelope))
+        let peer = check_unwrap find_existing_device(this.peer_id)
+
+        check is_valid_psk_length(this.psk_length_in_bytes);
```
gknopf-aranya, line 831:
```suggestion
        check is_valid_psk_length(this.psk_length_in_bytes)
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1687 @@
...
+              author.device_id == this.receiver_id
+
+        let peer_id = if author.device_id == this.sender_id {
+            :this.receiver_id
+        } else {
+            :this.sender_id
+        }
+        let peer = check_unwrap find_existing_device(peer_id)
+
+        check is_valid_psk_length(this.psk_length_in_bytes);
```
gknopf-aranya, line 1019:
```suggestion
        check is_valid_psk_length(this.psk_length_in_bytes)
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1687 @@
...
+        label_id: label_id,
+        peer_encap: ch.peer_encap,
+        psk_length_in_bytes: ch.psk_length_in_bytes,
+    }
+}
+
+// The effect that is emitted when the author of a unidirectional
+// AQC channel successfully processes the `AqcCreateUniChannel`
+// command.
+effect AqcUniChannelCreated {
```
gknopf-aranya, line 933: Should the `AqcUniChannelCreated` effect also have a `author_secrets_id` field like `AceBidiChannelCreated`?

elagergren-spideroak: yep, oversight, sending the fix

docs/quic-channels.md
```diff
@@ -0,0 +1,1715 @@
...
+    encap bytes,
+    // The size in bytes of the PSK.
+    //
+    // Per the AQC specification, this must be at least 32 and
+    // less than 2^16.
+    psk_length_in_bytes int,
+}
+
+command AqcCreateUniChannel {
+    fields {
```
gknopf-aranya, line 991: Missing the `psk_length_in_bytes` field needed by:
`check is_valid_psk_length(this.psk_length_in_bytes)`

docs/quic-channels.md
```diff
@@ -0,0 +1,1715 @@
...
+        let label_id = envelope::command_id(envelope)
+
+        // NB: Check roles, other ACLs here.
+
+        // Verify that the label does not already exist.
+        //
+        // This will happen in the `finish` block if we try to
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        check !exists Label[label_id: this.label_id]
```
gknopf-aranya, line 1193: `label_id` isn't an input of the `CreateLabel` command, it's generated by `let label_id = envelope::command_id(envelope)`. So `this.label_id` won't work here:
```suggestion
        check !exists Label[label_id: label_id]
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1715 @@
...
+
+        finish {
+            create Label[label_id: this.label_id]=>{name: this.label_name, author_id: author.device_id}
```
gknopf-aranya, line 1196: Ditto here:
```suggestion
            create Label[label_id: label_id]=>{name: this.label_name, author_id: author.device_id}
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1715 @@
...
+// Emits `QueriedLabelAssignment` for all labels the device has
+// been granted permission to use.
+action query_label_assignments(device_id id) {
+    map AssignedLabel[device_id: device_id, label_id: ?] as f {
+        publish QueryLabelAssignment {
+            device_id: device_id,
+            label_id: f.label_id,
+            label_name: f.name,
+            label_author_id: f.author_id,
+        }
```
gknopf-aranya, line 1463: The `AssignedLabel` fact doesn't have all the fields we're trying to return here.
We could query the label to get the extra fields or add more fields to the `AssignedLabel` fact.
```suggestion
        let label = check_unwrap query Label[label_id: f.label_id]
        publish QueryLabelAssignment {
            device_id: f.device_id,
            label_id: label.label_id,
            label_name: label.name,
            label_author_id: label.author_id,
        }
```

docs/quic-channels.md
elagergren-spideroak: yeah, just typos

docs/quic-channels.md
```diff
@@ -0,0 +1,1720 @@
...
+        // This will happen in the `finish` block if we try to
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        let label = check_unwrap query Label[label_id: this.label_id]
+
+        finish {
+            delete Label[label_id: label.label_id]=>{}
+
+            // Cascade the label assignments.
+            delete AssignedLabel[device_id: ?, label_id: label.label_id]
```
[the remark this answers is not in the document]
gknopf-aranya: Oh, could be because this is missing `author_id`:
```
            emit LabelDeleted {
                label_id: this.label_id,
                label_name: label.name,
                label_author_id: label.author_id,
            }
```

docs/quic-channels.md
```diff
@@ -0,0 +1,1720 @@
...
+
+            emit LabelDeleted {
+                label_id: this.label_id,
+                label_name: label.name,
+                label_author_id: label.author_id,
+            }
```
gknopf-aranya, line 1259:
```suggestion
            emit LabelDeleted {
                label_id: this.label_id,
                label_name: label.name,
                label_author_id: label.author_id,
                author_id: author.device_id,
            }
```

docs/quic-channels.md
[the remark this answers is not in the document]
gknopf-aranya: Resolving this and opening a different convo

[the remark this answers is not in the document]
jdygert-spok: Wildcard binds need to be last in any query including deletes. It is a bug that the compiler does not detect this.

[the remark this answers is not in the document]
gknopf-aranya: Policy checked into this PR isn't guaranteed to compile. I've copied the latest over to my PR here and have everything compiling:
https://github.com/aranya-project/aranya/pull/145/files#diff-50b0db5827a2e66fcb36996671bcdb2241f59d65b4bfe0ac96aa8f32840aec07

docs/quic-channels.md
```diff
@@ -0,0 +1,1721 @@
...
+struct Device {
+    // The device's ID.
+    device_id id,
+}
+
+// Returns a device.
+function get_valid_device(device_id id) struct Device
+
+// Returns the device's encoded public encryption key.
+function get_enc_pk(device_id id) bytes
```
mooreoak, line 612: Is this just stubbed out?

mooreoak (commented):
I don't see any issues with the design but we will need to fill out the stubbed policy before we can land this.

docs/quic-channels.md
elagergren-spideroak: Yes. This code already exists in Aranya's policy. It's stubbed out here b/c it would be silly and distracting to copy and paste all of Aranya's IDAM code into this PR.

elagergren-spideroak:
> I don't see any issues with the design but we will need to fill out the stubbed policy before we can land this.

Please see my comment about the stubbed out policy.

docs/quic-channels.md
```diff
@@ -0,0 +1,1721 @@
...
+Each channel will hold a connection `Handle` which can be used to open new
+streams. A channel can have any number of streams.
+
+```rust
+/// Identifies a unique channel between two peers.
+pub struct AqcChannelID {
+    /// The node id of the peer.
+    node_id: NodeId,
+    /// The channel label. This allows multiple channels between two peers.
+    label: Label,
```
gknopf-aranya, line 1573: Here and elsewhere, this should be updated to use the `LabelId` type since we will no longer be using `u32` integers for AQC labels.
```suggestion
    label_id: LabelId,
```

djohnson-spideroak (approved):

mooreoak (approved):

PR #34, opened 2025-01-22
Add walkthrough

### LINKED ISSUE #33: Move `walkthrough.md` to public docs
The walkthrough guide should be made available on our public docs site. Currently, it lives in the [aranya repo](https://github.com/aranya-project/aranya/blob/main/docs/walkthrough.md), but this makes it a bit obfuscated and it should be easier to find. 

The walkthrough should be removed from the aranya repo when it is moved over to the docs site.

shannynat:
First half of moving the walkthrough from the main aranya repo to the public docs site. 

Will add link to PR for removing the walkthrough when it is posted. 

Closes #33

### DIFF
--- a/docs/walkthrough.md
+++ b/docs/walkthrough.md
@@ -1,0 +1,721 @@
+---
+layout: page
+title: Aranya Walkthrough
+permalink: "/walkthrough/"
+---
+
+# Getting Started with Aranya
+
+Note: The following walkthrough offers a detailed explanation of Aranya's
+internals, written in Rust, to assist in setting up an example usage. To run
+this scenario using our C API wrappers, see the [C example](../examples/c/).
+
+In this document, we will walk through a scenario with five users initializing
+and running Aranya. The users will create a team using Aranya and send messages
+to each other using Aranya Fast Channels. There are a few things to keep in mind:
+
+- Any policy actions are determined by the implemented policy. This walkthrough
+  will use the default policy defined [here](/crates/aranya-daemon/src/policy.md).
+
+- Security tip: This walkthrough is intended as an example to be run on a
+  single machine. As such, a single machine is used to build all key bundles
+  and run all daemons under a single user's profile. In production, each Aranya
+  user's key bundle should be created under separate Linux users on their
+  respective machines and preferably all private keys should be stored in a
+  protected partition, such as an HSM, for maximum security. This avoids a
+  single access point for all Aranya user keys in case a machine is compromised.
+
+See the [examples](../examples) directory for example applications that
+follow scenarios similar to the one described in this document. Also, see the
+[overview documentation](overview.md) for more details on the components
+used in this walkthrough.
+
+# Outline
+
+The walkthrough includes five users who will be referred to by their user role.
+The actions performed by each user are based on the permissions assigned to
+each role in the [default policy](/crates/aranya-daemon/src/policy.md). There
+will be five users, `Owner`, `Admin`, `Operator`, `Member A` and `Member B`. We
+will use the [`daemon`](/crates/aranya-daemon/src/daemon.rs) implementation for
+this example.
+
+Step 1. [Prepare the device environment](#prereqs)
+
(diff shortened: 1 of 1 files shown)


docs/walkthrough.md
gknopf-aranya, line 1: @smckeon-so are you able to verify that this change looks good on the website? How can we verify website changes on PRs without modifying the production website on the `main` branch?

docs/walkthrough.md
```diff
@@ -0,0 +1,721 @@
...
+layout: page
+title: Aranya Walkthrough
+permalink: "/walkthrough/"
+---
+
+# Getting Started with Aranya
+
+Note: The following walkthrough offers a detailed explanation of Aranya's
+internals, written in Rust, to assist in setting up an example usage. To run
+this scenario using our C API wrappers, see the [C example](../examples/c/).
```
gknopf-aranya, line 11: URLs will need to be updated to point to the right repository. Relative links won't work if this link is on the website and the code is in a different repo.

shannynat: Oh, shoot forgot about those. Thanks for catching it!

docs/walkthrough.md
```diff
@@ -0,0 +1,721 @@
+---
+layout: page
+title: Aranya Walkthrough
+permalink: "/walkthrough/"
+---
+
+# Getting Started with Aranya
```
gknopf-aranya, line 7: Since this walkthrough is being moved to `aranya-docs`, we'll want to mention that the instructions are for the `aranya` repo. That used to be implied by the walkthrough being in the `aranya` repo. Since that's no longer the case, readers will need the additional context.

shannynat: Ok, I added a link to the aranya repo where `Aranya` is first mentioned, the first paragraph below this. Lmk if you think it should be more obvious

docs/walkthrough.md
shannynat: Updated the links!

docs/walkthrough.md
smckeon-so: Short answer is we can't. There's no staging server in this pipeline. As far as I know GH Pages doesn't support it. We can however run it locally on a jekyll dev server.

gknopf-aranya: We should document how to do this so it's easier for devs. Where would the best place be to document it? The README of this repo possibly?

smckeon-so: Here's a PR to add this to the readme. https://github.com/aranya-project/aranya-docs/pull/35

docs/walkthrough.md
ycarmy: It might help to be a little more specific by saying something about how this walkthrough uses the daemon integration of Aranya (include link to the repo) since you could then add a following sentence per Scott's comment https://github.com/aranya-project/aranya-docs/issues/33#issuecomment-2610641680 to reference where other material related to the library-only integration exists.

shannynat: Added a note that the walkthrough uses the aranya repo as a client/daemon integration (with links) and that the aranya-runtime crate can be used to integrate with aranya as a library only. Since we don't have documentation on the library-only integration, I just included a link to the runtime crate for the time being. Can add a story to link the docs when we have them. 

I also rearranged the information at the top (it felt a little weird and congested having multiple notes/things to keep in mind).

shannynat: Since we do not have the documentation (or clarity on whether we want to include information on integrating directly with aranya-core) _yet_, removed the reference to integrating with aranya-core/runtime. I did keep the information rearranging and a note about what the walkthrough uses in the examples.

ycarmy (approved):
Looks good!

gknopf-aranya (approved):

smckeon-so (approved):

docs/walkthrough.md
gknopf-aranya: @shannynat can this conversation be resolved?

PR #56, opened 2025-02-11
Add glossary

smckeon-so:
This PR adds the glossary from our internal wiki to the docs site. 

Closes #40

PR #65, opened 2025-02-14
54/Update user to device in specs and docs

### LINKED ISSUE #54: Specs and documentation should refer to "Device", not "User"
Go through the docs and replace any instance of "User" or "Entity" with "Device". We want to be consistent with the descriptions we use for devices/users/entities, currently we have a mix of the three.

smckeon-so:
This PR updates the naming of users/entities to devices. As part of the wider naming convention change to device, we're updating it here in the specs  and docs. 

Given that we haven't done the change in the wider codebases, we may want to hold off on merging this in, but I'd like some feedback as to if the wording was updated in the correct places. Namely where we talk about device roles and the role device. We had a role for a user, along with admin and operator, etc. Should this now be "admin, operator, device" in the policy examples?

Closes #54


nikki-aranya (dismissed):
This is definitely a welcome change, reads a lot cleaner now that we have one name to call things.

docs/aranya-beta.md
```diff
@@ -169,7 +169,7 @@ returns the device ID of the device that joined using that code.
 returns the team_id of the team that was joined.
 
 The goal of this onboarding API is to simplify the process of onboarding a
-user/device by providing an invite code instead of passing a KeyBundle.
```
shannynat, line 172: We have `user_key_bundle` (ln 152) and `user` (ln 264) as a role, I assume that's what the policy itself says, hence why we wouldn't change it? (Is there an aranya-beta policy?)

smckeon-so:
Until we can update user/entity in the code/policy, the plan should be to use device in the explainer text, but keep user/entity in the code examples. As we update aranya and aranya-core we can update the docs.

gknopf-aranya:
We are tracking naming consistency changes here:
https://github.com/aranya-project/aranya/issues/42

Work has not yet been started on changing that in the codebase.
But we are hoping to pick it up soon once the incremental release is complete.

gknopf-aranya:
Do we define what a device is anywhere? I think it's mentioned only in the MVP spec currently.
We may also want to add it to the glossary and discuss it elsewhere as applicable in the overview/getting-started docs.

gknopf-aranya (approved):
Changes look good to me. Left a comment about adding docs about what a device is to glossary/overview/getting-started docs.

smckeon-so:
> Do we define what a device is anywhere? I think it's mentioned only in the MVP spec currently. We may also want to add it to the glossary and discuss it elsewhere as applicable in the overview/getting-started docs.

@gknopf-aranya  We have a reference to Devicein the glossary PR. https://github.com/aranya-project/aranya-docs/pull/56

djohnson-spideroak (approved):

PR #81, opened 2025-03-11
mvp spec: query api updates

gknopf-aranya:
Started working on fact DB queries and noticed a few improvements we could make to the fact DB query API.

### DIFF
--- a/docs/aranya-mvp.md
+++ b/docs/aranya-mvp.md
@@ -260,12 +260,12 @@ FactDB queries over the current perspective of the graph should be possible thro
 commands in the policy that will return query results in their emitted effects. These APIs
 are likely to be moved to nice-to-have or Post-MVP, but are currently planned for MVP.
 
-- `QueryRoleAssignment(device_id) -> Role`
-- `QueryDeviceKeybundle(device_id) -> Keybundle`
-- `QueryAfcNetworkId(device_id) -> network_str`
-- `QueryAfcLabelAssignments(device_id) -> Vec<label>`
-- `QueryAfcLabelExists(device_id) -> Vec<label>`
-
+- `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
+- `QueryRoleAssignment(team_id, device_id) -> Role`
+- `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
+- `QueryAfcNetworkId(team_id, device_id) -> network_str`
+- `QueryAfcLabelExists(team_id, label) -> bool`
+- `QueryAfcLabelAssignments(team_id, device_id) -> Vec<label>`
 
 ## Roles & Permissions
 


docs/aranya-mvp.md
```diff
@@ -260,12 +260,12 @@ FactDB queries over the current perspective of the graph should be possible thro
 commands in the policy that will return query results in their emitted effects. These APIs
 are likely to be moved to nice-to-have or Post-MVP, but are currently planned for MVP.
 
-- `QueryRoleAssignment(device_id) -> Role`
-- `QueryDeviceKeybundle(device_id) -> Keybundle`
-- `QueryAfcNetworkId(device_id) -> network_str`
-- `QueryAfcLabelAssignments(device_id) -> Vec<label>`
-- `QueryAfcLabelExists(device_id) -> Vec<label>`
-
+- `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
```
gknopf-aranya, line 263: @djohnson-spideroak, @ycarmy I'd like to support this API method but I'm not sure if the policy supports returning a vector.

gknopf-aranya: In the Rust/C examples we can iterate over devices ids for devices on the team. In a distributed deployment, we could query the device ID from an endpoint such as tarpc. So I can work around not having a vec in the policy for now.

docs/aranya-mvp.md
```diff
@@ -260,12 +266,12 @@ FactDB queries over the current perspective of the graph should be possible thro
...
-- `QueryAfcNetworkId(device_id) -> network_str`
-- `QueryAfcLabelAssignments(device_id) -> Vec<label>`
-- `QueryAfcLabelExists(device_id) -> Vec<label>`
-
+- `QueryRoleAssignment(team_id, device_id) -> Role`
+- `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
+- `QueryAfcNetIdentifier(team_id, device_id) -> network_str`
+- `QueryAqcNetIdentifier(team_id, device_id) -> network_str`
+- `QueryLabelExists(team_id, label) -> bool`
+- `QueryLabelAssignments(team_id, device_id) -> Vec<label>`
```
gknopf-aranya, line 274: @djohnson-spideroak, @chip-so, @ycarmy the policy won't be able to return a vector of labels.

docs/aranya-mvp.md
```diff
@@ -260,12 +266,12 @@ FactDB queries over the current perspective of the graph should be possible thro
 commands in the policy that will return query results in their emitted effects. These APIs
 are likely to be moved to nice-to-have or Post-MVP, but are currently planned for MVP.
 
-- `QueryRoleAssignment(device_id) -> Role`
-- `QueryDeviceKeybundle(device_id) -> Keybundle`
-- `QueryAfcNetworkId(device_id) -> network_str`
-- `QueryAfcLabelAssignments(device_id) -> Vec<label>`
-- `QueryAfcLabelExists(device_id) -> Vec<label>`
-
+- `QueryRoleAssignment(team_id, device_id) -> Role`
```
gknopf-aranya, line 269:
```suggestion
- `QueryDeviceRole(team_id, device_id) -> Role`
```

gknopf-aranya, line 269: Do we want a method for querying whether a label is assigned to a specific device?

docs/aranya-mvp.md
gknopf-aranya: That could return multiple commands, but if the policy is unable to loop over all the labels in the factdb this won't help.

jdygert-spok: Oh right, you can only `map` in an action. This is quite inefficient but it does work. You map over the fact in the action, publishing a command for each label. Then each command emits an effect with that label. Thus your sink gets a bunch of effects, each containing one label.

gknopf-aranya: What would the syntax for that look like? Are there any existing policy examples of iteratively invoking `map`?

jdygert-spok: See the policy in this test: https://github.com/aranya-project/aranya-core/blob/acb26b2327d9effee5e95cb3be3dcf0de9301808/crates/aranya-policy-vm/tests/vm.rs#L2050

jdygert-spok: We'd either need a second fact or to swap `label/user_id` order if we aren't relying on it, since we can only do prefix queries.

gknopf-aranya: If we swap label/user_id, we could lookup all the `ChanOp` for each channel.
Perhaps we'd want to store the `label` in the fact in order to look that up.
Seems a little odd to have the `label` in the key and value, but if there isn't a way to tell what the current key is when iterating using map, then we'd need some other way to retrieve the key, which could be stored in the value.

docs/aranya-mvp.md
```diff
@@ -260,12 +266,13 @@ FactDB queries over the current perspective of the graph should be possible thro
...
-- `QueryAfcLabelAssignments(device_id) -> Vec<label>`
-- `QueryAfcLabelExists(device_id) -> Vec<label>`
-
+- `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
+- `QueryRoleAssignment(team_id, device_id) -> Role`
+- `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
+- `QueryAfcNetIdentifier(team_id, device_id) -> network_str`
+- `QueryAqcNetIdentifier(team_id, device_id) -> network_str`
+- `QueryLabelExists(team_id, label) -> bool`
+- `QueryLabelAssignments(team_id, device_id) -> Vec<label>`
```
gknopf-aranya, line 275:
```suggestion
- `QueryDeviceLabelAssignments(team_id, device_id) -> Vec<label>`
```

docs/aranya-mvp.md
```diff
@@ -260,12 +266,13 @@ FactDB queries over the current perspective of the graph should be possible thro
...
-- `QueryAfcLabelAssignments(device_id) -> Vec<label>`
-- `QueryAfcLabelExists(device_id) -> Vec<label>`
-
+- `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
+- `QueryDeviceRole(team_id, device_id) -> Role`
+- `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
+- `QueryAfcNetIdentifier(team_id, device_id) -> network_str`
+- `QueryAqcNetIdentifier(team_id, device_id) -> network_str`
+- `QueryLabelExists(team_id, label) -> bool`
+- `QueryDeviceLabelAssignments(team_id, device_id) -> Vec<label>`
```
gknopf-aranya, line 275:
```suggestion
- `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
- `QueryDeviceRole(team_id, device_id) -> Role`
- `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
- `QueryDeviceLabelAssignments(team_id, device_id) -> Vec<label>`
- `QueryAfcNetIdentifier(team_id, device_id) -> network_str`
- `QueryAqcNetIdentifier(team_id, device_id) -> network_str`
- `QueryLabelExists(team_id, label) -> bool`
```

docs/aranya-mvp.md
gknopf-aranya: @djohnson-spideroak, @ycarmy, @jdygert-spok

docs/aranya-mvp.md
jdygert-spok: The fact given in the loop will have both the keys and values.

docs/aranya-mvp.md
```diff
@@ -247,7 +247,13 @@ device for use with AFC. If the address already exists for this device, it is re
 address. Capable of resolving addresses via DNS. For use with CreateChannel and receiving messages.
 Can take either DNS name, IPv4, or IPv6. Current implementation uses a bidi map, so we can reverse
 lookup.
-- `UnsetAfcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a network address from a device.
+- `UnsetAfcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a AFC network address from a device.
+- `SetAqcNetIdentifier(team_id, device_id, net_identifier)` - associate a network address to a
+device for use with AQC. If the address already exists for this device, it is replaced with the new
+address. Capable of resolving addresses via DNS. For use with CreateChannel and receiving messages.
```
elagergren-spideroak, line 253: > Capable of resolving addresses via DNS.

What does this mean? The API itself will not perform DNS lookups. AQC (or AFC) will perform the DNS lookup, if necessary (e.g., no lookup is required for IP addresses).

docs/aranya-mvp.md
```diff
@@ -247,7 +247,13 @@ device for use with AFC. If the address already exists for this device, it is re
 address. Capable of resolving addresses via DNS. For use with CreateChannel and receiving messages.
 Can take either DNS name, IPv4, or IPv6. Current implementation uses a bidi map, so we can reverse
 lookup.
-- `UnsetAfcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a network address from a device.
+- `UnsetAfcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a AFC network address from a device.
+- `SetAqcNetIdentifier(team_id, device_id, net_identifier)` - associate a network address to a
```
elagergren-spideroak, line 251: For AQC and AFC we should _probably_ ensure that the device ID <-> net ID is 1:1.

However, this does _not_ preclude multiple DNS names from resolving to the same IP address (obviously, since that is the entire purpose of using DNS names).

docs/aranya-mvp.md
gknopf-aranya: This wording is from the original MVP spec. I believe it means that a network identifier can have a DNS lookup performed on it. cc: @djohnson-spideroak, @ycarmy.

docs/aranya-mvp.md
```diff
@@ -247,7 +247,13 @@ device for use with AFC. If the address already exists for this device, it is re
...
+device for use with AQC. If the address already exists for this device, it is replaced with the new
+address. Capable of resolving addresses via DNS. For use with CreateChannel and receiving messages.
+Can take either DNS name, IPv4, or IPv6. Current implementation uses a bidi map, so we can reverse
+lookup.
+- `UnsetAqcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a AQC network address from a device.
```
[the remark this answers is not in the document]
gknopf-aranya: Do you mean that after syncing the peers on the other side of the channel will destroy their side of the channel?
The user that called `UnsetAqcNetIdentifier` would know to destroy its channels locally.

docs/aranya-mvp.md
gknopf-aranya: We can filter through the labels returned from `QueryDeviceLabelAssignments`, so this is not needed.

docs/aranya-mvp.md
elagergren-spideroak: What I meant is that if device A has the net identifier "foo.com" then device B cannot _also_ have the net identifier "foo.com". IOW, net identifiers are globally unique within a team.

I kinda can see valid use cases for duplicate net identifiers, but until we have a concrete use case we should not support it in the MVP.

docs/aranya-mvp.md
elagergren-spideroak: Let's remove the wording. It's confusing at best and we already specify that a net identifier can be a DNS name, which implies that you can perform DNS lookups on it.

docs/aranya-mvp.md
```diff
@@ -244,10 +244,16 @@ flag to allow embedded devices and advanced users to access them.
 
 - `SetAfcNetIdentifier(team_id, device_id, net_identifier)` - associate a network address to a
 device for use with AFC. If the address already exists for this device, it is replaced with the new
-address. Capable of resolving addresses via DNS. For use with CreateChannel and receiving messages.
+address. For use with CreateChannel and receiving messages. There is a 1:1 mapping between a device ID and the network identifier used for AFC which means different devices cannot have the same network identifier.
 Can take either DNS name, IPv4, or IPv6. Current implementation uses a bidi map, so we can reverse
 lookup.
-- `UnsetAfcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a network address from a device.
+- `UnsetAfcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a AFC network address from a device. This prevents the device from creating new channels where it receives data from a peer until `SetAfcNetIdentifier` is called again and destroys all existing channels where the device can receive data.
```
djohnson-spideroak, line 250:
```suggestion
- `UnsetAfcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a AFC network address from a device. This prevents the device from creating new channels where it receives data from a peer until `SetAfcNetIdentifier` is called again. Destroys all existing channels where the device can receive data.
```

docs/aranya-mvp.md
```diff
@@ -244,10 +244,16 @@ flag to allow embedded devices and advanced users to access them.
...
+- `SetAqcNetIdentifier(team_id, device_id, net_identifier)` - associate a network address to a
+device for use with AQC. If the address already exists for this device, it is replaced with the new
+address. For use with CreateChannel and receiving messages. There is a 1:1 mapping between a device ID and the network identifier used for AQC which means different devices cannot have the same network identifier.
+Can take either DNS name, IPv4, or IPv6. Current implementation uses a bidi map, so we can reverse
+lookup.
+- `UnsetAqcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a AQC network address from a device. This prevents the device from creating new channels where it receives data from a peer until `SetAqcNetIdentifier` is called again and destroys all existing channels where the device can receive data.
```
djohnson-spideroak, line 256:
```suggestion
- `UnsetAqcNetIdentifier(team_id, device_id, net_identifier)` - disassociate a AQC network address from a device. This prevents the device from creating new channels where it receives data from a peer until `SetAqcNetIdentifier` is called again. Destroys all existing channels where the device can receive data.
```

docs/aranya-mvp.md
djohnson-spideroak: We would implement this via a map in an action, with a special effect that the daemon would store in an accumulator. Implementing this as a synchronous method would be tricky though.

djohnson-spideroak (approved):

gknopf-aranya:
@elagergren-spideroak, @jdygert-spok this is ready for review

PR #89, opened 2025-03-26
QUIC channels crypto

elagergren-spideroak:


docs/quic-channels.md
```diff
@@ -1,40 +1,432 @@
...
 
 ## Overview
 
-Aranya Quic Channels are a way for application data to be sent over the 
-network to other peers using TLS to ensure security. The QUIC protocol will
-be used to send data. This provides multiplexing of connections, TLS for
-security, and custom congestion control. We will default to using BBRv2
-for congestion control.
+Aranya QUIC Channels are end-to-end encrypted, topic-segmented
+communication channels between two devices. AQC channels are
```
gknopf-aranya, line 12: Each channel can establish communication between two devices:
```suggestion
An Aranya QUIC Channel is an end-to-end encrypted, topic-segmented
communication channel between two devices. AQC channels are
```

docs/quic-channels.md
```diff
@@ -1,40 +1,432 @@
...
+easily created or destroyed.
+
+Channels can be either bidirectional or unidirectional.
+Bidirectional channels allow both devices to send and receive
+data. Unidirectional channels allow one device to send data and
+the other device to receive data. Generally speaking,
+bidirectional channels are the default channel type.
+
+The two devices participating in a channel are referred to as the
+channel participants. The participant the created the channel is
```
gknopf-aranya, line 34:
```suggestion
channel participants. The participant that created the channel is
```

docs/quic-channels.md
```diff
@@ -1,40 +1,432 @@
...
+implemented on top of QUIC, which uses TLS for security, allows
+multiplexing connections, and custom congestion control
+(defaulting to BBRv2).
 
-## Connections
+## Design
+
+### Channels
+
+A channel is comprised of one or more QUIC connections between
```
gknopf-aranya, line 21: @elagergren-spideroak, @djohnson-spideroak, @benz-spideroak does a AQC channel support multiple connections or streams? I thought we were going to use streams since they are lighter weight.
A QUIC stream would reuse an existing QUIC connection to send data over a different stream.
A new QUIC connection would have another handshake and potentially different TLS key material.
Would each QUIC connection for a channel use the same TLS key material?

docs/quic-channels.md
```diff
@@ -1,40 +1,432 @@
...
+create a channel for device B and device C.
+
+### Labels
+
+Each channel is associated with a label, which defines the
+channel's topic.
+
+Devices with sufficient permissions (typically administrators)
+can create labels, and devices with sufficient permissions can
+grant other devices permission to use those labels. Devices can
```
gknopf-aranya, line 47: The code and policy refers to granting permissions to use a label as `assign_label` or `AssignLabel`. I think we should use the word "assign" in this spec for consistency.

docs/quic-channels.md
elagergren-spideroak: > does a AQC channel support multiple connections or streams?  [...]

An AQC channel is comprised of one or more QUIC connections.
A QUIC connection is comprised of one or more QUIC streams.

QUIC's cryptography (TLS 1.3) occurs at the connection level, so one QUIC connection _cannot_ support multiple channels.

> Would each QUIC connection for a channel use the same TLS key material?

Yes. This is described in detail elsewhere in the document.

docs/quic-channels.md
elagergren-spideroak: I understand that the policy uses "assign." I intentionally chose this verbiage.

"... grant other devices permission to use those labels" is significantly more clear than "... assigns those labels to other devices," especially in a security context. It also helps reinforce that multiple devices can use a label (i.e., that a label can be assigned to multiple devices).

However, `assign_label` is a nicer function name than `grant_device_permission_to_use_label`. The docs for `assign_label` _should_ say something like, "Grants the device permission to use the label." That's what I wrote when I originally wrote AFC's policy, anyway.

docs/quic-channels.md
gknopf-aranya: I wasn't asking if a QUIC connection can support multiple channels. It's clear that each AQC channel needs a different QUIC connection in order to use different TLS key material.
What I'm asking is if we are streaming different data streams between peers on an AQC channel will that open a new QUIC connection for the channel or open a new QUIC stream on the existing QUIC connection? It would make sense to open a QUIC stream since opening a new QUIC connection each time has additional overhead.

docs/quic-channels.md
```diff
@@ -1,40 +1,432 @@
...
+   participate in channels.
+5. Channels should allow participants to transmit a reasonable
+   amount of data without accidentally exceeding those limits.
+
+(1) and (2) are solved by [RFC 8446], [RFC 9000], and [HPKE].
+Additionally, trust in the device `EncryptionKey`s are rooted in
+the Aranya graph.
+
+(3) is solved by contextual binding. All channel parameters are
+includes as contextual binding, so if any channel parameters
```
gknopf-aranya, line 301:
```suggestion
(3) is solved by contextual binding. All channel parameters are
included as contextual binding, so if any channel parameters
```

docs/quic-channels.md
```diff
@@ -1,40 +1,432 @@
...
+
+The optimial bound is 2^r where the collision risk is 2^-r after
+`r` connections. Per [Abdalla and Bellare](rekey), this bound can
+be approximated as 2^(n/3).
+
+Both the client and server provide 256-bit nonces, so each
+connection effectively has a 512-bit nonce. However, we
+conservatively assume that only the client's nonce is truly
+random. This places the optimal bound at 2^(256/3) = 2^85
+connections. If the client creates QUIC conenctions at the
```
gknopf-aranya, line 324:
```suggestion
connections. If the client creates QUIC connections at the
```

docs/quic-channels.md
```diff
@@ -1,40 +1,432 @@
...
+  are unlikely to collide with resumption PSK identities.
+
+##### Privacy Considerations
+
+In general, most privacy considerations can be found in [RFC
+8446], [RFC 9000], [HPKE], and Aranya's specifications.
+
+###### PSK Identities
+
+If [Encrypted Client Hello](ECH) (ECH) is used, PSK identities
```
gknopf-aranya, line 398: Is ECH used by default? How would a user enable/disable using ECH?

mooreoak (approved):

docs/quic-channels.md
gknopf-aranya: It could be that whether a new QUIC connection or steam is opened is an implementation detail of AQC.
Although which one is selected does have cryptographic implications. Should we document the security implications of opening one or the other? E.g. if a stream is opened, it will use the same keys as the other streams in the channel. If a new QUIC connection is opened, it will use different keys derived from the PSK.

gknopf-aranya (approved):

docs/quic-channels.md
```diff
@@ -1,40 +1,437 @@
...
+        "ID-v1",
+        suite_id,
+        bytes(enc),
+        "UniChannelId",
+    )
+    return id
+}
+```
+
+#### Unidirectional Channel PSK
```
benz-spideroak, line 195: Are unidirectional channels needed? Once the connection is open bidirectional or unidirectional streams can be opened.

gknopf-aranya: My understanding of unidirectional channels is that the direction is cryptographically enforced.
If a unidirectional AQC channel is opened A -> B, if B sends to A, A will not be able to decrypt the message from B.
If A sends to B, B will be able to decrypt the message from A.
You could open a bidirectional or unidirectional stream in QUIC and send data either direction, but the AQC crypto determines read/write permissions, not the QUIC implementation.

gknopf-aranya: @benz-spideroak, @elagergren-spideroak, @mooreoak I think the AQC implementation should check the AQC channel type/direction and prevent sending data in the wrong direction for a unidirectional channel. There's no point sending data to a peer if we already know it won't be able to decrypt that data.

docs/quic-channels.md
```diff
@@ -1,40 +1,437 @@
...
+        bytes(enc),
+        "UniChannelId",
+    )
+    return id
+}
+```
+
+#### TLS Certificates
+
+This feature is currently schedule for after MVP.
```
nikki-aranya, line 282:
```suggestion
This feature is currently scheduled for after MVP.
```

nikki-aranya (approved):
You could probably rearrange the writeup since some things are mentioned before being introduced (e.g. what "encapsulation" is in PSK identities at the bottom, or max connections), but otherwise looks good

docs/quic-channels.md
elagergren-spideroak: > My understanding of unidirectional channels is that the direction is cryptographically enforced.

No. Both peers necessarily agree on the channel direction, but in both protocols both peers necessarily know all the keys. This isn't anything new to AQC, it's also part of AFC. It is just how symmetric cryptography works.

In AFC and AQC both peers agree on a shared secret (IKM) that is used to derive the actual secrets used in the protocols. In AFC, they derive two keys: A and B. One peer encrypts with A and decrypts with B while the other peer encrypts with B and decrypts with A. In AQC, they derive a PSK.

This means that in an AFC unidirectional channel the "receive only" peer could (physically) send encrypted data. But the "send only" peer would raise an error because it knows that the channel is unidirectional. Our implementation of AFC programmatically—not cryptographically—ensures that each peer respects the channel direction (see [`Directed`][0] and [`ChanDirection`][1]).

AQC should do the same thing. Perhaps via unidirectional streams?

For background: a sample use case for unidirectional channels is telemetry. In most cases telemetry is one way: a producer (a satellite, for example) needs to fire off its telemetry to a consumer (a ground station).

[0]: https://github.com/aranya-project/aranya-core/blob/d4bc63365844f6310440b361e8a4bca2171ffd1f/crates/aranya-fast-channels/src/state.rs#L267
[1]: https://github.com/aranya-project/aranya-core/blob/d4bc63365844f6310440b361e8a4bca2171ffd1f/crates/aranya-fast-channels/src/shm/shared.rs#L276

elagergren-spideroak: > I also don't see why this is a feature needed to the MVP.

I am okay with putting them behind an experimental feature flag if needed. But they have been part of AFC and AQC's feature sets from the beginning, so we can't remove them as a feature. We also have all of the rest of the machinery in place to support uni channels for AQC.

PR #96, opened 2025-05-01
aqc: devices must have permission to assign labels

elagergren-spideroak:
This *could* be in the IDAM or custom roles spec, but I think it's fine to put it here for now.

### DIFF
--- a/docs/quic-channels.md
+++ b/docs/quic-channels.md
@@ -599,10 +599,18 @@ use envelope
 use idam
 use perspective
 
+// NB: The following structs and functions are stubbed out.
+
 // A device.
 struct Device {
     // The device's ID.
     device_id id,
+}
+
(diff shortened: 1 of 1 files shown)


elagergren-spideroak:
Still a WIP.

/cc: @mooreoak @gknopf-aranya @ycarmy

docs/quic-channels.md
```diff
@@ -1354,8 +1478,16 @@ command AssignLabel {
 
         // NB: Check roles, other ACLs here.
 
+        // Devices are never allowed to assign roles to
+        // themselves.
+        check target.device_id != author.device_id
```
gknopf-aranya, line 1483: This is implicitly checked when validating `author.precedence > target.precedence` because a device cannot have a higher precedence than itself.

PR #107, opened 2025-06-23
Allow expressions in match blocks; remove braces for if expressions

### LINKED ISSUE #87: Update spec to allow expressions in `if`/`match` blocks
See comments in https://github.com/aranya-project/aranya-core/pull/119.

```
let x = match y {
  1 => {
    :y+1
  }
  _ => 0 // simple expressions are allowed by the compiler, but not spec only allows expression blocks.
}
```

chip-so:
Closes #87

### DIFF
--- a/docs/policy-v2.md
+++ b/docs/policy-v2.md
@@ -43,26 +43,11 @@ global scope is the last scope in this search and is always present.
 current scope. Shadowing an existing identifier this way is a compile
 error.
 
-## Block Expressions
+## Match Expressions
 
-A block expression contains zero or more statements followed by a
-required colon separator and terminal expression. The expression is the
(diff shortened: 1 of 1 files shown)


docs/policy-v2.md
```diff
@@ -84,36 +69,51 @@ action foo(x int) {
...
+        let v = ffi::external_value()
+        : x + v
+    }
+}
+```
+
+## `if` expression improvements
+
+`if` expressions have now removed the braces around expressions. More
+complex operations can be performed through block expressions.
```
gknopf-aranya, line 103: Can you show an example of handling an if/else with a block expression?

gknopf-aranya (approved):

PR #113, opened 2025-07-16
Passive sync spec

benz-spideroak:


docs/bidi-sync.md
```diff
@@ -0,0 +1,295 @@
...
+    category: docs
+---
+
+# Bidirectional Sync
+
+## Overview
+
+Bidirectional sync allows two Aranya peers to synchronize with each other in both directions using a single QUIC stream connection. This is particularly useful in scenarios where network connections may be asymmetric - where one peer can easily initiate connections to another, but the reverse direction is difficult or impossible.
+
+In traditional unidirectional sync, Peer A connects to Peer B, sends a sync request, receives a sync response, and closes the connection. If Peer B also needs to sync with Peer A, it must initiate a separate connection. With bidirectional sync, after Peer A receives Peer B's sync response, Peer B can immediately send its own sync request over the same stream, allowing both peers to synchronize in a single connection.
```
gknopf-aranya, line 13: The QUIC syncer uses streams for sync requests. So each sync request doesn't need to be a separate connection.
```suggestion
In traditional unidirectional sync, Peer A connects to Peer B, sends a sync request, receives a sync response, and closes the corresponding QUIC stream. A traditional unidirectional sync is a 2-way handshake. If Peer B also needs to sync with Peer A, it must initiate a separate sync request handshake in the other direction. With bidirectional sync, after Peer A receives Peer B's sync response, Peer B can immediately send its own sync request over the same stream, then Peer A responds with a sync response to PeerB, allowing both peers to synchronize in a single 3-way handshake.
```

docs/bidi-sync.md
```diff
@@ -0,0 +1,295 @@
...
+
+The `sync_impl` method in the QUIC `State` is modified to handle bidirectional sync:
+
+```rust
+async fn sync_impl<S>(
+    syncer: &mut Syncer<Self>,
+    id: GraphId,
+    sink: &mut S,
+    peer: &Addr,
+) -> impl Future<Output = SyncResult<()>> + Send
```
gknopf-aranya, line 140: `SyncPeerConfig` isn't used directly in `task.rs`. There is a delay queue that stores a `SyncPeer` type which can have a `bidirectional: bool` field added to it. The `bidirectional` bool can be taken in as a function parameter here:
```suggestion
async fn sync_impl<S>(
    syncer: &mut Syncer<Self>,
    id: GraphId,
    sink: &mut S,
    peer: &Addr,
    bidirectional: bool,
) -> impl Future<Output = SyncResult<()>> + Send
```
Then `sync_impl()` can be called from `sync()` like this:
`if let Err(e) = <ST as SyncState>::sync_impl(self, peer.graph_id, &mut sink, &peer.addr, &peer.bidirectional)`

docs/bidi-sync.md
```diff
@@ -0,0 +1,295 @@
...
+        // Send initial sync request
+        syncer.send_sync_request(&mut send, &mut sync_requester, peer, bidirectional).await?;
+
+        // Receive sync response  
+        syncer.receive_sync_response(&mut recv, &mut sync_requester, &id, sink, peer).await?;
+
+        // If bidirectional, handle server's sync request
+        if bidirectional {
+            syncer.handle_server_sync_request(&mut recv, &mut send, &id, sink, peer).await?;
+        }
```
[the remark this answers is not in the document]
gknopf-aranya: It seems more clear to show the request and response explicitly at the protocol level rather than hiding both operations inside a single handler method.

docs/bidi-sync.md
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+```
+
+#### Default Values
+
+- `bidirectional` defaults to `false` to maintain backward compatibility
+- Existing configurations continue to work unchanged
+
+## Backward Compatibility
```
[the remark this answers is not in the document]
ssweetney-spok: Do we even have a version a bump. The current quic syncer is using [this](https://github.com/aranya-project/aranya/blob/d74f88812f9183cf844665528da926ce0f526d20/crates/aranya-daemon/src/sync/task/quic.rs#L66-L67) for the APLN

docs/bidi-sync.md
```diff
@@ -0,0 +1,295 @@
...
+
+The existing `SyncType::Poll` variant is extended to include a bidirectional flag:
+
+```rust
+pub enum SyncType {
+    Poll {
+        request: SyncRequestMessage,
+        address: ServerAddress,
+        // NEW: Indicates whether this is a bidirectional sync request
+        bidirectional: bool,
```
gknopf-aranya, line 27: Can this new `bidirectional` field be added to `SyncRequestMessage` instead so when the server receives the sync request it can parse `SyncRequestMessage` to determine whether to make a sync request in the other direction?

docs/bidi-sync.md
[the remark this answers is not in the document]
gknopf-aranya: If we want to maintain backward compatibility with `quic-sync-unstable`, I think we'd want to fallback to unidirectional sync if that version is observed. But, my understanding is that if the ALPN protocols don't match, the QUIC connection cannot be established. This means if we don't change the `quic-sync-unstable` ALPN version when adding bidirectional sync, incompatible versions of the syncer could be used together. If we do bump the ALPN version, old/new syncers could no longer communicate with each other even though the change to support uni/bidi sync in a backward compatible way is trivial.
I think we may have made a mistake by removing the version byte from the sync protocol since we no longer have a way to bump the protocol and syncer versions independently.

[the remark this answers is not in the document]
gknopf-aranya: Having a single protocol version means that there's no way to bump the protocol version in a backward compatible way. Even if we implemented the protocol in a backward compatible way, bumping the protocol version would mean newer protocol version can no longer communicate with older versions.
Agree that supporting backward compatibility between 1.0 and 2.0 isn't needed right now because we're already planning to make a breaking changes with the PSK -> cert change.
But we will need to be able to make backward compatible changes to 2.0+ if 2.0 is committing to API stability.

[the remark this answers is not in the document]
benz-spideroak: By backwards compatible I just mean that you won't have to change code that calls sync. Yes both peers will need to run the new version to enable bidi sync.

docs/bidi-sync.md
benz-spideroak: Why is that better? The server will still receive it if it's here.

gknopf-aranya: It depends on if we want bidirectional sync to be supported by the sync protocol in `aranya-core` or only by our specific implementation of the sync protocol in `aranya`.

docs/bidi-sync.md
benz-spideroak: What's a 3-way handshake?

gknopf-aranya: 3-way handshake because the bidirectional sync protocol completes after 3 phases: A -> B, A <- B, A -> B.
TCP connection is an example of a 3-way handshake.

docs/bidi-sync.md
```diff
@@ -0,0 +1,295 @@
...
+```
+Client                    Server
+  |                         |
+  |--- SyncRequest -------->|
+  |    (bidirectional=true) |
+  |                         |
+  |<--- SyncResponse -------|
+  |                         |
+  |<--- SyncRequest --------|
+  |    (from server)        |
```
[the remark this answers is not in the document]
djohnson-spideroak: This is a good fallback for a quick implementation. Ideally, I think we would like to try and reduce the amount of redundant data. If bidirectional=true, we can probably just expect the client to immediately send a sync response. I think there are a lot more details though around actually calculating what commands we would need to send.

docs/bidi-sync.md
```diff
@@ -0,0 +1,279 @@
...
+## Testing
+
+New test cases should cover:
+
+1. **Basic Bidirectional Flow**: Verify both directions complete successfully
+2. **Backward Compatibility**: Ensure unidirectional sync continues to work unchanged
+3. **Configuration**: Test all combinations of `bidirectional` flag settings
+
+## Future Considerations
+
```
djohnson-spideroak, line 279: Future considerations are keeping the stream/connection open and doing syncs on an interval. it may affect how we implement this issue.

docs/bidi-sync.md
djohnson-spideroak: I am also not sure that the QUIC protocol implementation details belong here, this is more about the sync protocol.

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+
+**Scenario**: Both peers can connect to each other.
+
+**Configuration**:
+- Both peers: `passive = false`
+
+**Flow**:
+1. First peer to need sync establishes the connection
+2. Both peers use this connection for all subsequent sync operations
+3. Connection remains persistent for ongoing sync needs
```
ssweetney-spok, line 297: I didn't include the ability for the syncer and sync server to share connections in the current implementation. It seems like that change would have to be made in order for this flow to work as expected

[Relevant thread](https://github.com/aranya-project/aranya/pull/258/files/bad144f4eae20f9da54ba56b1facadfb518594bc#r2082168019)

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+
+// Periodic sync - respects passive flag in config
+async fn add_sync_peer(addr: Addr, team: TeamId, config: SyncPeerConfig) -> Result<()>;
+```
+
+When `passive` is enabled:
+- `sync_now`: Only syncs if a connection to the peer already exists, otherwise does nothing
+- `add_sync_peer`: Periodic sync attempts only occur when a connection exists
+
+If `cfg` is `None` for `sync_now`, default behavior is used (can initiate connections).
```
djohnson-spideroak, line 91: I think we also need an API option to open the connection but not sync.

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+}
+
+// In QUIC State implementation
+impl SyncState for State {
+    async fn sync_impl<S>(
+        syncer: &mut Syncer<Self>,
+        id: GraphId,
+        sink: &mut S,
+        peer: &Addr,
+        config: &SyncPeerConfig, // NEW parameter
```
djohnson-spideroak, line 216: Do we want to pipe the configs fully down into the sync impl? Is there a reason we weren't doing this before?

ssweetney-spok: We weren't doing it because currently there are no values in the config that directly affect the sync implementation. "interval" and "sync_now" affect _when_ a sync happens but now _how_. The new passive field would be the first

docs/passive-sync.md
gknopf-aranya: I'm thinking we could add another field to the sync config to control whether a connection is opened without syncing:
```
SyncPeerConfig {
    connect: bool,
    ...
}
```
Seems simpler to do this than to add a new API method just for opening a QUIC connection.

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+Currently, peers maintain persistent QUIC connections and open individual streams for sync operations. However, if a peer wants to sync and no connection exists, it will always attempt to open a new connection. Passive sync adds the ability to configure peers to only sync through existing connections, providing control over connection initiation in asymmetric network environments.
+
+## Design
+
+### Core Concepts
+
+1. **Single Persistent Connection**: Only one QUIC connection exists between any two peers at a time
+2. **Either Peer Can Connect**: The connection can be initiated by either peer 
+3. **Stream-per-Sync**: Each sync operation opens its own stream on the shared connection
+4. **Connection Control**: Peers can be configured to only sync through existing connections (not initiate new ones)
```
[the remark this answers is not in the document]
gknopf-aranya: The change that we're making is allowing a peer to initiate a QUIC connection without syncing so that either peer can open the persistent connection. Once the connection has been opened, each peer can optionally initiate syncs in either direction.

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+2. **Connection Lost**: Connection drops during sync operation
+3. **Stream Failure**: Individual sync stream fails while connection remains healthy
+
+A new error type is introduced:
+
+```rust
+#[derive(Error, Debug)]
+pub enum PassiveSyncError {
+    #[error("Sync requires existing connection but none found for peer {0}")]
+    NoConnectionAvailable(Addr),
```
gknopf-aranya, line 120: This error will only be returned if the connection doesn't exist yet and the client is unable to create it

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+### Configuration
+
+#### SyncPeerConfig Extension
+
+The `SyncPeerConfig` struct is extended with a connection-only flag:
+
+```rust
+pub struct SyncPeerConfig {
```
gknopf-aranya, line 67: `SyncPeerConfig` should be generalized for different types of syncers that may be added in the future. So any new fields added specifically for the QUIC syncer should be optional.

docs/passive-sync.md
gknopf-aranya, line 1: I'm not convinced we need a passive sync mode. The idea is to allow either peer to initiate the connection.
It doesn't hurt for both peers to attempt to initiate the connection and would make the syncer implementation more resilient if that is allowed. There is a small cost to attempting to initiate a connection on a network that won't allow it, but the increase in resilience that provides may be worth it.

docs/passive-sync.md
benz-spideroak: Why would we need to be able to open a connection, but not sync? Is it as a performance optimization to not have the sync? Or do you think it's more clear as a separate API method?

It seems like premature optimization to me. We know we need passive sync because of network configuration, but it feels like a lot of complexity to add a set of APIs to manage connections when they may not be needed.

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+## Security Considerations
+
+Passive sync maintains security properties:
+
+1. Persistent connections use the same PSK-secured QUIC as before
+2. Each stream follows existing sync protocol authentication
+3. Connection management doesn't bypass existing security validation
+4. Stream isolation prevents cross-contamination between sync operations
+
+## Testing
```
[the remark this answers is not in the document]
benz-spideroak: What does that mean?

docs/passive-sync.md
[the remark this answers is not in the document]
gknopf-aranya: From my notes on the Rust call:
"We don't really need bidirectional sync. We just need a way to initiate the QUIC connection from either side."

docs/passive-sync.md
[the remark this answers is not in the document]
gknopf-aranya: B opens connection without syncing.
A syncs with B using existing connection (rather than creating a new connection).

docs/passive-sync.md
```diff
@@ -0,0 +1,314 @@
...
+#### SyncPeerConfig Extension
+
+The `SyncPeerConfig` struct is extended with a connection-only flag:
+
+```rust
+pub struct SyncPeerConfig {
+    pub interval: Duration,
+    pub sync_now: bool,
+    // NEW: If true, only sync through existing connections (don't initiate new ones)
+    pub passive: bool,
```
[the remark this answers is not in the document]
gknopf-aranya: When the bidirectional sync issue was opened, there was an older version of the syncer that would open a new QUIC connection for each sync.
This meant that in order to sync through a firewall, the connection would be opened from one side and send a sync request in both directions through a stream before closing the connection.
The current syncer implementation reuses connections that have been opened previously, so as long as both sides of the connection sync, at least one side will hopefully establish a QUIC connection. That means there is less of a need to support opening a connection from either side than there used to be, because that is already supported by the way the syncer is implemented. A potential optimization is to allow a peer to open a connection without syncing.

PR #119, opened 2025-07-28
118: Sync Hello Spec

benz-spideroak:


docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+    // Unsubscribe from hello notifications
+    Unsubscribe {
+        team_id: TeamId,
+    },
+    // Notification message sent to subscribers
+    Hello {
+        /// The team ID this hello applies to
+        team_id: TeamId,
+        /// The current head of the sender's graph
+        head_id: Address,
```
gknopf-aranya, line 46: Could set this to `None` if the head is unchanged since the last hello message.
```suggestion
        head_id: Option<Address>,
```

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+1. **Trigger**: A hello message is sent to subscribers whenever the local graph head changes due to:
+   - Local action execution that adds new commands
+   - Sync operations that incorporate new commands from peers
+
+2. **Delay Management**: Messages are sent to each subscriber respecting their configured delay:
+   - `delay_seconds: 0` - notify immediately
+   - `delay_seconds: 1` - wait 1 second between notifying each subscriber
+   - Higher values provide longer delays between notifications
+
+3. **Delivery**: Hello messages are sent as fire-and-forget notifications - no acknowledgment or retry mechanism is required
```
gknopf-aranya, line 72: Acknowledgment and retries are built into the QUIC protocol. So if the hello messages are sent via QUIC, delivery would be guaranteed as long as there is connectivity between the peers.

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Receiving Hello Messages
+
+1. **Head Comparison**: Upon receiving a hello message, the peer compares the received head address with their own graph state
+
+2. **Sync Decision**: If the received head is unknown to the local graph, the peer may initiate a sync operation with the sender
```
gknopf-aranya, line 78: It may be easier for the peer to make a decision if a command sample is sent in addition to the head of the graph. Just sending the head command seems fine for the first iteration of this feature.

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+    Hello(SyncHelloType),
+}
+```
+
+### Connection Management
+
+Sync hello messages will reuse existing QUIC connections managed by the QUIC syncer:
+
+- **Connection Reuse**: Hello messages will leverage the existing `conns: BTreeMap<ConnectionKey, Connection>` maintained by the QUIC syncer state
+- **Connection Establishment**: If no open connection exists for a peer when sending a hello message, the syncer will attempt to establish a new QUIC connection using the same mechanism as regular sync operations
```
gknopf-aranya, line 120: This means the subscriber will likely open the original QUIC connection while the sender uses the existing connection.
If the connection ever closes and the sender doesn't have a way to reopen the connection (possibly due to a firewall), the subscriber wouldn't know to open a new connection if it is waiting to receive a hello message first.
Should the subscriber periodically check if there is an open connection for each subscription and reopen any that have closed?

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+    // New sync hello message type
+    Hello(SyncHelloType),
+}
+```
+
+### Connection Management
+
+Sync hello messages will reuse existing QUIC connections managed by the QUIC syncer:
+
+- **Connection Reuse**: Hello messages will leverage the existing `conns: BTreeMap<ConnectionKey, Connection>` maintained by the QUIC syncer state
```
gknopf-aranya, line 119: Subscriptions and hello messages will be sent on their own QUIC streams

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+    /// The team ID they're subscribed to
+    team_id: TeamId,
+    /// Delay in seconds between notifications to this subscriber
+    delay_seconds: u64,
+    /// Last notification time for delay management
+    last_notified: Option<Instant>,
+}
+
+// Storage: Map from (team_id, subscriber_address) to subscription details
+type HelloSubscriptions = HashMap<(TeamId, SocketAddr), HelloSubscription>;
```
gknopf-aranya, line 140: We'll want to maintain a map of subscriptions and subscribers.
Subscribers determine who to send hello messages to.
Subscriptions determine which subscriptions can be cancelled and whether delays are being respected.

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+}
+```
+
+## Operation
+
+### Subscription Management
+
+Peers must explicitly subscribe to receive hello notifications from other peers:
+
+1. **Subscribe**: Send a `Subscribe` message specifying the team ID and desired delay between notifications
```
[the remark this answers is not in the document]
djohnson-spideroak: Yes, I think unsolicited messages should be dropped.

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+2. **Subscription Storage**: The receiving peer stores subscription information including the subscriber's address, team ID, and delay preference
+3. **Unsubscribe**: Send an `Unsubscribe` message to stop receiving notifications
+
+### Sending Hello Messages
+
+1. **Trigger**: A hello message is sent to subscribers whenever the local graph head changes due to:
+   - Local action execution that adds new commands
```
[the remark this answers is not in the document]
djohnson-spideroak: We could configure a batching window, but for the first implementation we can defer batching until later.

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+
+// Storage: Map from (team_id, subscriber_address) to subscription details
+type HelloSubscriptions = HashMap<(TeamId, SocketAddr), HelloSubscription>;
+```
+
+### Delay Management
+
+To manage notification timing:
+
+- **Delay Scheduling**: Track `last_notified` time for each subscriber and respect their `delay_seconds` setting
```
djohnson-spideroak, line 147: rm. we do want some sort of rate limiting to prevent spam, but this doesnt really fit with how Moore and I initially designed this. The goal of the original delay was to allow devices to stagger notifications and prevent a flood of incoming sync requests if needed by adding a delay. I guess we need detail as well on how we would handle sending new notifications while the old updates are iterating between the peers for the previous notification.

docs/sync-hello.md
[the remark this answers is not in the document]
gknopf-aranya: We generally want the recipient to be in control of how much data is sent and how frequently it is sent so it can avoid DDOS'ing itself.

docs/sync-hello.md
```diff
@@ -0,0 +1,153 @@
...
+
+2. **Push syncing** (as described in [Open Sync Requests](/docs/open-sync-requests/)) can result in sending many duplicate commands to peers. When multiple peers push the same updates, recipients may receive the same command data repeatedly, wasting bandwidth and processing resources.
+
+We need a lightweight notification mechanism that alerts peers when new commands are available without sending the actual command data. This allows peers to control when they want to sync while avoiding both inefficient polling and duplicate command transmission.
+
+Sync hello provides this notification approach, letting peers know when updates are available so they can decide whether to initiate a sync operation.
+
+## Design
+
+Sync hello implements a subscription-based notification approach where peers can subscribe to receive lightweight "hello" messages containing the current graph head whenever the sender's graph is updated. Receiving peers can compare this head with their own state to determine if they need to sync.
```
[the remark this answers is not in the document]
benz-spideroak: The head of the graph will always be a single command. We merge until that's the case.

docs/sync-hello.md
benz-spideroak: Yes. Everything we send over a QUIC connection is sent over it's own stream.

docs/sync-hello.md
benz-spideroak: I don't see how that would be useful. The caller will need to track subscriptions and call an API to cancel them.

gknopf-aranya: It's useful for the client to be able to drop hello messages received earlier than expected based on the configured subscription delay.

docs/sync-hello.md
gknopf-aranya: The QUIC protocol itself guarantees delivery by resending dropped packets.
Once data has been sent and flushed, the send side of the stream can be closed and the peer is guaranteed to have received the data (as long as there were no connection/stream errors returned).
The connection object would need to stay open until the receiver gets the data.
https://docs.rs/s2n-quic/latest/s2n_quic/stream/struct.BidirectionalStream.html#method.flush

benz-spideroak: So for this case we're sending the hello message and ensuring that it arrived. But, because there's no acknowledgement we don't know if they did anything with it.

docs/sync-hello.md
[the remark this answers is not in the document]
benz-spideroak: We discussed this on the rust call. We'll have both. In the daemon config you can specify how often you want to send hello messages. And when a peer subscribes they can set how often they want to receive messages.

docs/sync-hello.md
```diff
@@ -0,0 +1,214 @@
...
+
+Additionally, peers can be configured to send hello notifications to specified peers without requiring those peers to subscribe to hello notifications first. This configuration-based approach allows peer A to automatically notify peer B based on local settings.
```
djohnson-spideroak, line 25: I am not sure we would want this, in general we want permission or an affirmative request to subscribe a peer.

docs/sync-hello.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Message Types
+
+```rust
+pub enum SyncHelloType {
+    // Subscribe to receive hello notifications from this peer
+    Subscribe {
+        /// The team ID to subscribe to updates for
+        team_id: TeamId,
+        /// Delay in seconds between notifications to this subscriber
```
djohnson-spideroak, line 35:
```suggestion
        /// Minimum delay in milliseconds between notifications to this subscriber
```

We should probably do milliseconds, not a huge change.

docs/sync-hello.md
benz-spideroak: This came out of a discussion on the rust call with Moore. He called this broadcasting.

gknopf-aranya (approved):

gknopf-aranya (approved):

PR #131, opened 2025-09-10
Policy book

chip-so:
This aims to collect and update the reference material in policy lang specs v1 and v2 into a nicely formatted mdbook website. To test it, install [mdBook](https://github.com/rust-lang/mdBook) (`cargo install mdbook`), then go into the `policy-book` directory and run `mdbook serve`.

### DIFF
--- a/policy-book/src/reference/statements/match.md
+++ b/policy-book/src/reference/statements/match.md
@@ -36,8 +36,8 @@ means `_` should be the last match arm, as nothing will be matched
 afterwards.
 
 A match expression must match exactly one arm in order for the match
-statement to be valid. Non-exhaustive matches may produce a compile
-error or a runtime exception when all match arms have failed to match.
+statement to be valid. Non-exhaustive matches will produce a compile
+error when all match arms have failed to match.
 Duplicate match arms will produce a compile error.
 
 Note that unlike the [`match`


policy-book/src/reference/basic-syntax.md
```diff
@@ -0,0 +1,42 @@
...
+
+Whitespace is not significant in Aranya Policy Language. Whitespace is
+any sequence of spaces, tabs, and newlines (which includes `\n`, `\r\n`,
+and `\r`).
+
+<img src="basic-syntax-whitespace.svg">
+
+## Comments
+
+Comments are C99-style, supporting both block comments(`/* */`) and
```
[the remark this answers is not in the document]
mooreoak: That is not surprising. It should be documented thou.

policy-book/src/reference/errors.md
```diff
@@ -0,0 +1,72 @@
...
+Runtime exceptions happen when an execution invariant is violated. Many
+things can cause runtime exceptions, including but not limited to:
+
+- `unwrap`ping `None`
+- integer over/underflow
+- Running out of memory (including overflowing the VM stack)
+- Creating a fact that already exists
+- VM stack underflow caused by compiler errors or badly behaving FFI
+
+There is no way to detect or recover from a runtime exception in the
```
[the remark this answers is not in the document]
mooreoak: They should not call the default recall block. If you hit and exception processing a command that has been added to the graph the system should stop opperating.

policy-book/src/reference/expressions.md
```diff
@@ -0,0 +1,5 @@
+# Expressions
+
+Expressions are similar to C-like languages, and have similar
+precedence. Operations on the wrong type will produce either a compile
+error or a runtime exception.
```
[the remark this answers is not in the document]
mooreoak: I guess I don't understand how this is all put together just reading the PR

policy-book/src/reference/expressions/operators.md
```diff
@@ -0,0 +1,70 @@
...
+| `\|\|`   | `A \|\| B` is true if A or B are true |
+| `!`      | `!A` is the logical negation of A |
+
+## Struct Operators
+
+| Operator | Meaning |
+|----------|---------|
+| `.`      | `A.B` accesses field B in struct A |
+| `as` | `A as B` creates a `struct B` from the fields of `A` only if the two struct types have the same fields |
+| `substruct` | `A substruct B` creates a `struct B` from the fields of `A` |
```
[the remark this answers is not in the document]
mooreoak: O so what is the current requirement?

policy-book/src/reference/statements/map.md
```diff
@@ -21,8 +21,9 @@ map FooCounter[deviceID: ?]=>{count: ?} as counter {
 ```
 
 `map` executes a fact query, and for each fact found, defines the given
-name as a struct containing the bound fields. This can result in zero
-or more iterations.
+name as a struct containing the fact's fields. This can result in zero
+or more iterations. Just like in `query`, the value side of the fact can
+be omitted if all its fields are bound.
```
mooreoak, line 26: Do you mean `all its fields are un-bound` or I might not understand the way we use bound.

policy-book/src/reference/top-level/actions.md
```diff
@@ -9,7 +9,13 @@ action foo(a int, b string) {
...
 An action is a function callable from the application, which can perform
 data transformations and publish zero or more commands. The effects of
-an action are atomic &ndash; the commands published and side effects
-emitted will only be visible to the rest of the system if the entire
-action succeeds. And error that causes termination will result in no
-changes (see [Errors in Actions](../errors.md)).
\ No newline at end of file
+an action are all or none &ndash; the commands published and side
+effects emitted will only be visible to the rest of the system if the
+entire action succeeds[^action-publish-clarification]. And error that
```
mooreoak, line 14: "And error" -> "An error"

mooreoak (changes requested):

policy-book/src/reference/expressions.md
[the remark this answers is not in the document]
recluse-so: @mooreoak and @chip-so is this resolved?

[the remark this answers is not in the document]
mooreoak: @recluse-so yes

policy-book/src/concepts/facts.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+Now which branch comes first determines whether the `Authorize` command
+in `D` is true or false. Because if we order it `A, C, B, D` it's like
+before, but if it's `A, B, D, C` we only have one `Increment` before the
+`Authorize` and it will produce `AuthorizationResult { authorized: false
+}`[^weave-order].
+
+[^weave-order]: Weave order isn't actually a branch-to-branch choice.
+    The resulting order could have interleaved `C` between `B` and `D`.
```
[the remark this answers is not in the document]
jdygert-spok: I think you updated this, but yes, to be clear, a command will always come after its ancestors.

policy-book/src/reference/basic-syntax.md
gknopf-aranya, line 1: There are images not loading on this page when I try to view it locally:
http://localhost:3000/reference/basic-syntax.html

gknopf-aranya: This was resolved after generating the railroad diagrams with `./build-diagrams.sh`

gknopf-aranya: Combining the diagram generation and serve steps into a single script would decrease the amount of manual steps required to serve `mdbook` locally. `serve.bash`:
```
./build-diagrams.bash
mdbook serve
```

policy-book/src/concepts/scope.md
```diff
@@ -0,0 +1,57 @@
+# Scope
+
+## Global Scope
+
+Variables can be defined in global scope, which makes them available to
```
gknopf-aranya, line 5: Do we have a use-case for defining anything other than constants in the global scope?
In general, it is considered bad practice to define mutable variables in a global scope.

gknopf-aranya:
Any reason not to host the rest of the specs and possibly the rest of our docs website with `mdbook` and do away with the Jekyll site?

policy-book/src/concepts/facts.md
```diff
@@ -0,0 +1,277 @@
...
+        check !exists Account[user: this.user]
+
+        finish {
+            create Account[user: this.user]=>{balance: 0}
+        }
+    }
+}
+```
+
+This command first [`check`](../reference/statements/check.md)s that no
```
gknopf-aranya, line 41: The `s` at the end of `checks` renders larger for me because it is not part of the link. Looks a bit strange aesthetically. This would look better if the entire word is the link.

policy-book/src/concepts/facts.md
```diff
@@ -0,0 +1,277 @@
...
+
+```policy
+query Users[user: this.user]=>{level: ?}
+```
+
+But trying to do the opposite &ndash; finding a user with the given
+authorization level &ndash; won't work as you expect.
+
+```policy
+query Users[user: ?]=>{level: AuthorizationLevel::Admin}
```
gknopf-aranya, line 255: Will this query fail if the first user doesn't have `level: AuthorizationLevel::Admin`? Or would the query continue looking until there is a match for the `level` value field?

policy-book/src/reference/basic-syntax.md
```diff
@@ -0,0 +1,42 @@
...
+line comments (`//`). Block comments do not nest.
+
+```policy
+/* This is a block comment
+   It can span multiple lines
```
gknopf-aranya, line 18: There are multiple places in our current policy that would benefit from using block comments instead of line comments cc: @jdygert-spok, @djohnson-spideroak

mooreoak: How do you think they would help? It seems that most people are against using them as they do not nest.

gknopf-aranya: There are multiple locations in the existing policy that use multi-line comments when they could be using block comments. I like block comments because I don't need to type an extra `//` on each line.

policy-book/src/reference/types/optional.md
```diff
@@ -0,0 +1,26 @@
+# Optional Types
+
+A type which can contain a value (`Some`) or no value (`None`). The type
```
gknopf-aranya, line 3:
```suggestion
A type which can contain a value (`Some(value)`) or no value (`None`). The type
```

policy-book/src/concepts/facts.md
jdygert-spok: I thought we had disallowed this at some point but I guess not. The current behavior is scanning over all matching facts by key, and checking for matching values.

policy-book/src/reference/types/structs.md
```diff
@@ -0,0 +1,246 @@
+# Structs
+
+A struct is a collection of named values accessed with the `.` operator.
```
gknopf-aranya, line 3:
```suggestion
A struct is a collection of named value fields accessed with the `.` operator.
```

policy-book/src/concepts/facts.md
gknopf-aranya: So this would return the entire set of user facts where `level: AuthorizationLevel::Admin`?

gknopf-aranya:
When this PR is merged, will the website be hosted publicly? Or is this expected to be a locally generated site?

policy-book/src/concepts/facts.md
jdygert-spok: `query` only ever returns 0 or 1 value. Currently this returns the first user that has `level: Admin`.

gknopf-aranya: If that's the case, I'm not sure the paragraph below is correct:
`It does not mean "find the first user whose level is admin".`

jdygert-spok: Yes, the behavior is wrong.

policy-book/src/reference/expressions/operators.md
```diff
@@ -0,0 +1,70 @@
...
+| `\|\|`   | `A \|\| B` is true if A or B are true |
+| `!`      | `!A` is the logical negation of A |
+
+## Struct Operators
+
+| Operator | Meaning |
+|----------|---------|
+| `.`      | `A.B` accesses field B in struct A |
+| `as` | `A as B` creates a `struct B` from the fields of `A` only if the two struct types have the same fields |
+| `substruct` | `A substruct B` creates a `struct B` from the fields of `A` only if the fields of `struct B` are a subset of `A`|
```
gknopf-aranya, line 46: Wondering why we need `as` and `substruct` keywords. We can use `substruct` in all cases where `as` is used, including cases where one of the structs has additional fields. Is the purpose to have an operation that will fail to compile unless both structs have exactly the same fields? As a policy writer, why do I care which method is used to convert between structs as long as the conversion is valid?

policy-book/src/reference/top-level/commands.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+The policy writer should not assume any particular structure or
+manipulate it directly.
+
+## Policy block
+
+The `policy` block contains statements which query data and check its
+validity. The `policy` block must terminate with a `finish` block
+(though it can have multiple `finish` blocks in branching paths). The
```
gknopf-aranya, line 137: Is it a compilation error to have multiple `finish` blocks on the same branch?

policy-book/src/reference/top-level/functions.md
```diff
@@ -0,0 +1,51 @@
...
+
+```
+finish function set_foo(device id, new int) {
+    update FooCounter[device: device]=>{count: current} to {count: new}
+}
+```
+
+Finish functions allow the abstraction of statements used in [`finish`
+blocks](commands.md#finish-block). Thus, they can only contain `finish`
+block statements (`create`, `update`, `delete`, `emit`) and cannot
```
gknopf-aranya, line 37: Would there be a performance benefit to implementing finish functions as finish macros?
Seems like all the `create`, `update`, `delete`, `emit` operations could be copied directly into the `finish` block at compile time.

policy-book/src/reference/top-level/ffi.md
```diff
@@ -0,0 +1,34 @@
+# Foreign Function Interface
+
+```
+use crypto
+use perspective
+```
+
+`use` defines an imported function library, which is provided by the
+application. You must call functions from imported libraries with the
```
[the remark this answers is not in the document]
gknopf-aranya: Do we have documentation on how a developer can implement a new ffi from scratch?

policy-book/src/reference/statements/return.md
```diff
@@ -0,0 +1,25 @@
...
+</div>
+
+```
+function foo() {
+    let x = unwrap query FooCount[deviceID: myId]
+    return x.count
+}
+```
+
+`return` evaluates an expression and returns the value from the
```
gknopf-aranya, line 24: Is it worth mentioning here that only built-in types or structs defined by the policy can be returned?

gknopf-aranya:
I'm not seeing docs on which parts of the policy will end up in the generated policy interface.

policy-book/src/reference/statements/return.md
gknopf-aranya: As opposed to some other type defined in a Rust crate or ffi.

policy-book/src/concepts/commands-graph.md
```diff
@@ -0,0 +1,112 @@
...
+    ...
+}
+```
+
+There are some more required parts here (see [the
+reference](../reference/top-level/commands.md)), but we'll omit them for
+brevity. First, the command has `fields`, which define the data stored
+in the command. And next it has `policy`, which is run both when the
+command is created in the local graph and when it is synced to other
+devices' graphs. It checks the validity of the data, then finally
```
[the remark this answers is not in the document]
chip-so: I think that makes sense, sure.

policy-book/src/reference/expressions/functions/if-match.md
```diff
@@ -0,0 +1,36 @@
...
+Both cases must have the same type or it is a compiler error.
+
+Not to be confused with the [`if` statement](../../statements/if.md).
+
+## `match`
+
+Much like the [`match` statement](../../statements/match.md), a `match`
+expression evaluates to one of its arm expressions based on an equality
+test against an expression. If no expression matches, it matches a
+default case represented by `_`. The default case must exist.
```
ssweetney-spok: That doesn't seem to be the case. We do check for [duplicate match arms](https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-policy-compiler/src/compile.rs#L2541-L2551) but the program will [panic if there's no default case and the scrutinee doesn't match anything](https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-policy-compiler/src/compile.rs#L2612-L2618)

chip-so:
> There doesn't appear to be documentation of the recently added `ephemeral` keyword

#142

policy-book/src/reference/statements/delete.md
```diff
@@ -0,0 +1,46 @@
...
+
+Prefix deletion is not yet supported in the policy VM. So a request to delete all matching records like:
+
+```
+delete FooCounter[deviceId: ?]
+```
+
+Will not delete all facts as expected.
+
+</div>
```
djohnson-spideroak: lets make a note in the prefix key deletion issue to update this section.

jnewell-spok (dismissed):
Looks good to me.

djohnson-spideroak (dismissed):

policy-book/src/reference/statements/match.md
```diff
@@ -0,0 +1,46 @@
...
+constant in each match arm. Except in the case of a `match`, all
+possibilities must be checked for.
+
+The `_` token is a "default" match that matches anything. This also
+means `_` should be the last match arm, as nothing will be matched
+afterwards.
+
+A match expression must match exactly one arm in order for the match
+statement to be valid. Non-exhaustive matches may produce a compile
+error or a runtime exception when all match arms have failed to match.
```
ssweetney-spok:
```suggestion
statement to be valid. Non-exhaustive matches will produce a compile
error when all match arms have failed to match.
```

ssweetney-spok (approved):

policy-book/src/concepts/facts.md
gknopf-aranya: Not sure. I recall a recent Rust call where a decision was made but don't recall what the decision was without referencing some notes.

chip-so: I'm 99% sure we agreed it was intended to work as documented and it needed to be fixed in the implementation. I'm just not finding a ticket for it.

chip-so: Anyway, issue here: aranya-project/aranya-core#488.

gknopf-aranya (approved):

djohnson-spideroak (approved):

PR #150, opened 2025-11-20
Result type spec

### LINKED ISSUE #149: Result type spec
Figure out the details of how to implement a Result-like return type.

apetkov-so:


docs/specs/result-type.md
```diff
@@ -0,0 +1,127 @@
+# Result Type Specification
+
+## Overview
+
+A `Result` type represents the outcome of an operation that can either succeed with a value, or fail with an error. This is a common pattern for error handling that makes error cases explicit and forces callers to handle them.
+
+## Type Definition
+
+```policy
+result <ok-type>, <err-type>
```
gknopf-aranya: Is this valid policy code? Since it's in a policy block, I'm expecting it to be valid policy code.

docs/specs/result-type.md
```diff
@@ -0,0 +1,127 @@
...
+let n = unwrap Err(-1) // policy check-exits
+```
+>TBD Probably shouldn't be allowed in policy/recall blocks
+
+#### `is_ok(result) bool`
+
+Returns `true` if the result is `Ok`.
+
+```policy
+if is_ok(result) {
```
gknopf-aranya:
```suggestion
if is_ok(Ok(42)) {
```

docs/specs/result-type.md
```diff
@@ -0,0 +1,127 @@
...
+    // proceed with success path
+}
+```
+
+#### `is_err(result) bool`
+
+Returns `true` if the result is `Err`.
+
+```policy
+if is_err(result) {
```
gknopf-aranya:
```suggestion
if is_err(Err(err)) {
```

docs/specs/result-type.md
```diff
@@ -0,0 +1,127 @@
...
+
+Results are typically consumed using pattern matching:
+
+```policy
+let n = match result {
+    Ok(value) => {
+        :value
+    }
+    Err(error) => {
+        return error // NOTE returning from an expression isn't supported
```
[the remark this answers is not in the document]
apetkov-so: It's not quite an example, yet. We're just trying to figure out the syntax, and uncover issues/todos.

docs/specs/result-type.md
```diff
@@ -0,0 +1,127 @@
...
+    ...
+    return Err(-1)
+}
+
+command Foo {
+    ...
+    policy {
+        let f = foo()
+
+        check is_ok(f), recall
```
gknopf-aranya: Would this have the same effect?
```suggestion
        f?
```

docs/specs/result-type.md
gknopf-aranya: True, I figured the example may be more clear if it's shown with a concrete type rather than a variable name representing an unknown expression.

docs/specs/result-type.md
```diff
@@ -0,0 +1,127 @@
...
+let value = foo() // the Return instruction adds a must_use obligation, which can be met with the following:
+let v = match value {
+    Err(e) => { return  e } // currently, we can't return from match expression
+    Ok(v) => :v
+}
+// this must appear immediately after the Const instruction that pushed the result
+// can't use the name `value`, though, because . 
+```
+
+This can be sugared with `?`, e.g. `let value = foo()?`.
```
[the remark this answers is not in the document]
mooreoak: I think we should implement `check_unwrap` as a short hand for this and solve the issue of what recall block to call and how.

docs/specs/result-type.md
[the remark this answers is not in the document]
mooreoak: I feel like with `is_ok()`, `is_error()`,  `unwrap`, and `check_unwrap` we could probably make this useful without the pattern matching. (I think the pattern matching is great I just don't think we *need* it for this feature to be useful.)

mooreoak (commented):
This looks good overall. My biggest suggestion is that we only implement `is_ok()`, `is_error()`, `unwrap`, and `check_unwrap`, in the first pass and not `?` and matching. The `?` and matching are great but I think the are beyond what we need for mvp.

docs/specs/result-type.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+### Matching result values
+
+Result values can be used in match statements or expressions. The patterns are:
+```
+Ok(ident) => ...
+Err(ident) => ...
+```
+Where `ident` holds the success or error value, respectively.
```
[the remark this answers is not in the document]
apetkov-so: #574 fixes that and is close to landing. I'll wait until it does, and update the spec.

apetkov-so:
Results are implemented and documented in the policy book, so we don't need the spec anymore.

PR #159, opened 2025-12-15
Add policy runner spec

chip-so:
Ported from https://github.com/aranya-project/aranya-core/pull/521. Annoyingly, I have made some updates and you'll have to read it again.

### DIFF
--- a/docs/policy-runner.md
+++ b/docs/policy-runner.md
@@ -74,7 +74,8 @@ The additional values section defines values that will be made available
 to policy execution as globals (and thus available in the action calls
 and command structs described above). This will be able to define any
 legal `Value`s supported by the policy VM, including `id` and `bytes`
-values that the policy cannot express as literals.
+values that the policy cannot express as literals. Additionally, it will
+be able to generate keys which will be stored in the keystore.
 
 ## Operation
 
@@ -85,13 +86,17 @@ command-line arguments.
 policy-tester [OPTIONS] <POLICY> <RUNFILE> [RUNFILE ...]
 ```
 
-First it will compile the policy file into a VM. Then for each run file,
-it will load the file, define any additional values specified in it,
-then execute its sequence of actions and command structs. Additional
-values defined in earlier run files will remain defined while executing
-later run files, to allow sequences to be composed through multiple
-files. For example, you can have one run file that sets up a team, then
-several others which perform more specific team manipulations.
+First, the run files are all parsed. Additional values in these files
+are combined into additional global definitions. Next, it will compile
+the policy file into a VM including the additional global definitions.
+Then for each run file, it will execute its sequence of actions and
+command structs, outputting effects as they are committed to the sink.
+
+All additional values are defined while executing run files, which means
+their identifiers cannot overlap. This also allows sequences to be
+composed through multiple files. For example, you can have one run file
+that sets up a team, then several others which perform more specific
+team manipulations using the keys defined in that earlier team run file.
 
 Any effects produced will be printed out in the `Display` implementation
 for `Value`s, which describes every fields' content, including `id` and


docs/policy-tester.md
```diff
@@ -0,0 +1,91 @@
...
+`crypto`, `idam`, and `afc` FFIs depend on a key store, and they will
+use the implementation specified above.
+
+The `device` FFI needs a device ID specified. One will be randomly
+generated and stored in `<wd>/device_id`.
+
+## Test Library
+
+This is probably a useful set of tools to have generally for integration
+testing, so the non-CLI parts of this should be available as a library
```
gknopf-aranya: Does this mean the policy tester library will be made available as a crate in `aranya-core`?

docs/policy-tester.md
```diff
@@ -0,0 +1,91 @@
...
+globals. This section will be able to define `id` and `bytes` values
+that the policy cannot.
+
+And finally, there will be an optional validation section that contains
+a sequence of effect structs that are expected to be produced.
+
+## Operation
+
+The tool will accept a policy file and a test file as command-line
+arguments. First it will compile the policy file into a VM. Then it will
```
gknopf-aranya: It seems like a common use-case would be to run tests on an entire test suite.
Providing the test tool with a working directly that points to all the policy and test files in the test suite would allow it to run multiple tests on different policies.

chip-so: I could just make it accept multiple test files and run them in sequence. Just run it as e.g. `policy-tester policy.md tests/*`.

docs/policy-tester.md
```diff
@@ -0,0 +1,91 @@
...
+load the test file, define any additional values specified in it, then
+execute its sequence of actions and command structs.
+
+If the validation section is present, the effects emitted by execution
+will be checked against it. The tool will alert the user if there is a
+deviation, and produce a diff between the expected and received effects.
+If the validation section is absent or overridden by a command-line
```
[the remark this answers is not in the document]
chip-so: As described in the previous section, it is a sequence of effect structs. Exact match only in this version.

docs/policy-tester.md
chip-so: Yeah. And if you want to test multiple files, you can pretty easily loop that yourself.

```
for P in policies/*.md; do T=$(basename $P .md); policy-tester $P tests/$T/*; done
```

docs/policy-tester.md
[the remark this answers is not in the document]
ssweetney-spok: > So I will need some basic "match anything" designator when specifying effects, and some extended policy format

For the initial implementation, could we use struct composition with a special named value like "received" to handle fields with types like `id` and `bytes`?

chip-so:
Reworked this to be a "runner" rather than any kind of test framework. Also realized that `spideroak-crypto` does not provide any manually-seeded RNG, so for the seeded option I'll have to create an RNG implementation (though I should just be able to steal some of the internals from `spideroak-crypto`).

docs/policy-runner.md
```diff
@@ -0,0 +1,96 @@
...
+The tool will accept a policy file and one or more run files as
+command-line arguments.
+
+```
+policy-tester [OPTIONS] <POLICY> <RUNFILE> [RUNFILE ...]
+```
+
+First it will compile the policy file into a VM. Then for each run file,
+it will load the file, define any additional values specified in it,
+then execute its sequence of actions and command structs. Additional
```
gknopf-aranya: If we allow multiple test files in the input, we'll want a way to delineate their respective effect sequences in the output.

docs/policy-runner.md
```diff
@@ -0,0 +1,96 @@
+# Aranya Policy Tester
+
+The Aranya Policy Tester is a CLI tool to execute actions and commands
```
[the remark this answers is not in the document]
gknopf-aranya: I think it's relevant information for whoever is going to read the spec and implement the feature. It may not be strictly part of the spec, design, or requirements. But, I tend to think losing the original rationale for why something is designed/implemented a certain way can result in missing the mark in the final implementation. It's useful for certain design/implementation tradeoffs someone might need to make which aren't fully specified from the start.

docs/policy-runner.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Crypto Engine
+
+The "default" crypto engine (`aranya_crypto::default:DefaultEngine`)
+will be used. The root key used by the engine will be randomly generated
+and stored under `<wd>/crypto_root_key`.
+
+### Key Store
+
+The tool will use the `fs_keystore` implementation with data stored
```
ssweetney-spok: nit: I think it's supposed to refer to `aranya_crypto::keystore::fs_keystore`? Some items are fully qualified but some aren't

chip-so: That's a good point. I'll clarify.

docs/policy-runner.md
```diff
@@ -0,0 +1,99 @@
...
+
+The tool will accept a policy file and one or more run files as
+command-line arguments.
+
+```
+policy-tester [OPTIONS] <POLICY> <RUNFILE> [RUNFILE ...]
+```
+
+First it will compile the policy file into a VM. Then for each run file,
+it will load the file, define any additional values specified in it,
```
chip-so: I think due to limitations I don't want to fix in the compiler/VM, this will actually define all the values from all run files at the start, which is a little different than I've written here. I'll update this.

docs/policy-runner.md
```diff
@@ -0,0 +1,104 @@
...
+`aranya-runtime`'s `VmPolicy` with all production FFIs - `envelope`,
+`perspective`, `device`, `crypto`, `idam`, and `afc`. Custom FFIs are
+not supported, but may be offered later through customizing the tester
+and recompiling.
+
+### Working Directory
+
+Run state for the crypto engine, keystore, and graph storage will be
+stored in a "working directory" configurable from the command-line. If
+no working directory is specified, the current directory is used. All
```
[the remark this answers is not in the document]
chip-so: The run files aren't related to the working directory. They're specified on the command-line. They can be anywhere.

docs/policy-runner.md
```diff
@@ -0,0 +1,104 @@
...
+policy itself (or the `vm_action!()` macro). These calls and structs
+will be compiled in the same way as the rest of the policy, which means
+they can use policy-defined types and global values.
+
+The additional values section defines values that will be made available
+to policy execution as globals (and thus available in the action calls
+and command structs described above). This will be able to define any
+legal `Value`s supported by the policy VM, including `id` and `bytes`
+values that the policy cannot express as literals. Additionally, it will
+be able to generate keys which will be stored in the keystore.
```
gknopf-aranya: The daemon stores keys in a persistent keystore on disk because they are needed again later when the daemon restarts. If policy testing runs are independent, is there a need for any persistent storage other than the input/output files? It may be worth considering an in-memory keystore for this.

gknopf-aranya (approved):

docs/policy-runner.md
chip-so: The runs are not defined to be independent. It stores state in the working directory explicitly to allow you to do multiple runs dependent on the same state. It might be useful to have an option to store nothing (or just do that if no working directory is specified), but I'd like to defer that to a later version as it will complicate the implementation.

chip-so: (It also stores data on disk because I believe it might be useful to run other tools that can visualize system state)

PR #160, opened 2025-12-19
mtls spec

gknopf-aranya:
Specification for using mTLS X.509 certs generated from an external PKI with the QUIC syncer.

### DIFF
--- a/docs/mtls.md
+++ b/docs/mtls.md
@@ -55,12 +55,14 @@ Future enhancements:
 - Must be able to generate a root CA cert and P-256 ECDSA secret key pair
 - Must be able to generate new certs signed by a root CA along with their P-256 ECDSA secret key
 - Must output certs in PEM format
+- Must allow a CN (Common Name) to be specified for each generated cert
 - Must allow SANs to be specified for non-CA certs such as DNS names and IP addresses
+- Must allow a validity period to be specified so certs can expire
 
 Example usage:
 ```bash
 # Create a root CA
-aranya-certgen ca --cert ca.pem --key ca.key --ca-name "My Company CA"
+aranya-certgen ca --cert ca.pem --key ca.key --ca-name "My Company CA" --validity-days 10
 
 # Create a signed certificate
 aranya-certgen signed \
@@ -69,6 +71,7 @@ aranya-certgen signed \
   --cn server \
   --dns example.com --dns www.example.com \
   --ip 192.168.1.10
+  --validity-days 10
 ```
 
 Future enhancements:


docs/mtls.md
```diff
@@ -0,0 +1,78 @@
...
+## Certificate Exchange
+
+mTLS certificates should be exchanged between peers prior to attempting to sync with each other. Each peer needs to be in possession of the mTLS x509 cert of the other peer before mutual mTLS authentication can occur.
+
+Certificates can be exchanged in one of two ways:
+- Directly, by exchanging certificates with a peer via a secure, out-of-band channel
+- Indirectly, by syncing certificates via the graph from another peer
+
+Add new Aranya APIs for adding/removing peer certs:
+- `AddPeerCert(DeviceId, x509_cert)` - adds a peer device's cert to the certificate store. The QUIC transport will now trust this cert. TODO: `NetworkIdentifier` for cert hostname?
```
gknopf-aranya: Should mention that trust extends to any certs signed by trusted certs.

docs/mtls.md
```diff
@@ -0,0 +1,111 @@
...
+In order to sync with each other, peers must either exchange their self-signed certs ahead of time, or posses a shared root cert that can be used to validate cert signatures.
+
+It is important to exchange certs via a secure out-of-band channel or they are susceptible to man-in-the-middle attacks.
+
+Certs can be exchanged in a few different ways:
+- Directly, by exchanging certs with a peer via a secure, out-of-band channel
+- Indirectly, by syncing certs via the graph from another peer
+- Indirectly, by sharing root certs which can be used to validate cert signatures
+
+Add new Aranya APIs for adding/removing/signing device certs:
```
mooreoak: I think we don't need any of these. We do need to configure the device with it's certs but we should accept those certs from external sources. That is today we should let customers leverage there existing PKI instead of implementing our own.

It seems reasonable to to have some bash scripts that wrap BoringSSL to generate certs and a CA for testing or customers that don't have existing PKI but we should not assume ownership of the PKI for the mTLS

docs/mtls.md
```diff
@@ -0,0 +1,111 @@
...
+
+Storing mTLS certs on the graph is beneficial, because it allows devices to sync the certs of other devices via the graph.
+
+However, there is a bootstrapping problem with this. If all the certs are stored on the graph, how does a peer initially sync the graph from another peer without knowing what mTLS certs to authenticate the QUIC transport with?
+
+The solution to this is to exchange x509 certs with a peer prior to the initial team sync with that peer. Once this initial sync completes, the device will have access to other certs via the graph that it can use to sync with other peers in the future.
+
+The initial team sync will provide the device with the team owner's mTLS cert and the certs of any devices that have already been added to the team. If any new devices are added to the team after that, the device would need to sync with one of the devices it already has a cert for, or obtain a cert directly from the peer it wishes to sync with.
+
+#### Adding/Removing Certs Via Graph Commands
```
mooreoak: I don't think we should store certs on the graph. Once we have the MVP working I think it would be good to store cert *revocations* on the graph but we would need to think carefully to make sure this dose not create any issues.

gknopf-aranya: The purpose of storing certs on the graph is to make it easier to sync certs a device doesn't have yet without needing to manually exchange them between peers. If root certs are specified during onboarding, does mean storing certs on the graph would no longer be needed?

docs/mtls.md
```diff
@@ -0,0 +1,111 @@
...
+
+Abbreviations in this document:
+- certificate -> cert
+- certificates -> certs
+
+## Certificate Generation
+
+When each peer generates its initial key bundle, it will also generate a random mTLS secret key from which the mTLS cert is derived.
+
+The mTLS secret key is a 2048 byte RSA key generated using Aranya's CSPRNG.
```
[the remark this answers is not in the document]
gknopf-aranya: What is the benefit of using ECDH vs. RSA?

docs/mtls.md
```diff
@@ -0,0 +1,111 @@
...
+A public, self-signed x509 mTLS cert is derived from the secret RSA key.
+
+It is recommended to use another device's cert (such as the team owner) as a root cert which signs device certs.
```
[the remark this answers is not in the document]
gknopf-aranya: Is the CA part of Aranya or an external CA not part of Aranya?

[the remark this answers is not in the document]
gknopf-aranya: Pushed a commit with external PKI

docs/mtls.md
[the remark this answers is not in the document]
gknopf-aranya: Looked it up. Better security with a smaller key size.

gknopf-aranya:
@mooreoak this spec has been updated for external PKI

gknopf-aranya:
@mooreoak, @djohnson-spideroak  mTLS spec has been updated with a daemon config for loading root/device certs when the daemon starts up

docs/mtls.md
```diff
@@ -0,0 +1,92 @@
...
+- The configured root certs and device cert are used for all QUIC connections and Aranya teams
+- QUIC connection attempts by the syncer should fail to be established if certs have not been configured/signed properly
+- QUIC connection attempts with expired certs should fail
+- Existing QUIC connections with expired certs should be closed
+- Security events such as failed authentication or signature verification should be logged
+
+Future enhancements:
+- Different root and device certs for different teams
+- Use system root certs
+- Verify that device cert is signed by one of the root certs when daemon loads, rather than failing later during TLS authentication
```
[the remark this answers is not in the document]
gknopf-aranya: From what I can tell, this verification is possible with the `openssl` and `rustls` crates.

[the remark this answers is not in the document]
gknopf-aranya: After discussing this with @mooreoak, we'll keep this earlier cert chain verification as a todo for the future.

docs/mtls.md
```diff
@@ -24,34 +24,40 @@ Abbreviations in this document:
...
     `<daemon_working_directory>/certs/device/`
         `cert.pem`
         `cert.key`
-- A single device cert is configured when the daemon loads. The device cert must be signed by one of the root certs.
+- A single device cert is configured when the daemon loads. The device cert must be signed by one of the root certs or an intermediate CA cert.
 - A set of root certs is configured when the Aranya daemon loads
 - The configured root certs and device cert are used for all QUIC connections and Aranya teams
+- QUIC connection attempts by the syncer should fail to be established if certs have not been configured/signed properly
+- QUIC connection attempts with expired certs should fail
+- Existing QUIC connections with expired certs should be closed
```
mooreoak: can we do this?

mooreoak (approved):

docs/mtls.md
```diff
@@ -24,34 +24,40 @@ Abbreviations in this document:
 
 ## Requirements
 
+- Use TLS >= 1.3 in the QUIC library (Rust QUIC libraries such as `s2n_quic` and `quinn` require TLS 1.3)
```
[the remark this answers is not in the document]
gknopf-aranya: I believe QUIC does require TLS 1.3. Although it might not be obvious to someone not very familiar with QUIC. Stating this explicitly helps to demonstrate the security properties of our mTLS implementation on top of QUIC.

docs/mtls.md
gknopf-aranya: Not sure. I'd need to read up on the QUIC lib docs more to verify whether it's possible.

gknopf-aranya: After looking into this more, I believe the QUIC protocol only performs the TLS handshake once when establishing the QUIC connection. So it wouldn't notice if certs expired after that initial handshake.

docs/mtls.md
```diff
@@ -24,34 +24,40 @@ Abbreviations in this document:
...
 mTLS root and device certs are generated externally via a user's existing PKI infrastructure.
-Device certs are signed by one of the root certs using the PKI infrastructure.
-An example CA that generates root and device certs will be provided for users that do not have an existing PKI infrastructure.
+Device certs are signed by one of the root certs or an intermediate CA cert using the PKI infrastructure.
 
-We recommend using ECDSA certs generated from a secret key of at least 224 bits.
+We recommend using P-256 ECDSA certs generated from a secret key of at least 256 bits.
 
-Aranya assumes that all certs have been generated and signed prior to configuring and loading the Aranya daemon.
+An example CA that generates root and device certs will be provided in an example application for users that do not have an existing PKI infrastructure.
```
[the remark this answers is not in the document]
gknopf-aranya: I believe the `openssl` cli tool can do this already. We can provide an example of how to generate/sign certs with `openssl`.

docs/mtls.md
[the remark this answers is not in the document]
gknopf-aranya: I've changed this to a note about TLS 1.3 being an implied requirement.

docs/mtls.md
gknopf-aranya: Removed this requirement

docs/mtls.md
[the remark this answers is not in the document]
gknopf-aranya: Updated this to mention providing an example of using the `openssl` tool to generate/sign certs.

docs/mtls.md
```diff
@@ -0,0 +1,93 @@
...
+- The configured root certs and device cert are used for all QUIC connections and Aranya teams
+- QUIC connection attempts by the syncer should fail to be established if certs have not been configured/signed properly
+- QUIC connection attempts with expired certs should fail
+- Security events such as failed authentication or signature verification should be logged
+
+Note:
+QUIC requires TLS 1.3 so that is an implied requirement. It's worth mentioning here since it is relevant to the security properties of our mTLS implementation.
+
+Future enhancements:
+- Different root and device certs for different teams
```
ssweetney-spok: Does using the same root and device certs for different teams pose security issues?

I think we'd also want to mention this in the future release notes because it's a regression from the current auth approach that uses different PSKs per team.

ssweetney-spok (approved):
LGTM

docs/mtls.md
```diff
@@ -0,0 +1,96 @@
...
+Because of this, we plan to migrate the authentication/encryption of the Aranya syncer transport from QUIC PSKs to QUIC mTLS certs.
+
+Abbreviations in this document:
+- certificate -> cert
+- certificates -> certs
+
+## Requirements
+
+- Users must be able to leverage their existing external PKI for generating/signing certs
+- mTLS certs must be X.509 TLS certs. We recommend using P-256 ECDSA secret keys of at least 256 bits to meet current NIST standards.
```
djohnson-spideroak: Lets add a source here to the NIST standard we are referencing.

docs/mtls.md
```diff
@@ -0,0 +1,96 @@
...
+root_certs=<root certs directory>
+device_cert=<device cert directory>
+...
+```
+
+Aranya assumes the following before loading certs into the QUIC syncer:
+- Certs have been generated and signed by the user's external PKI infrastructure
+- Certs include any relevant SANs information
+- The root certs directory contains at least one root certificate (e.g. `root1.pem`).
+- The device cert directory contains the device cert (e.g. `cert.pem`) signed by one of the root certs or an intermediate CA cert and the corresponding secret key (e.g. `cert.key`).
```
djohnson-spideroak: Lets add a note about what should happen if these are not met, even if its just "aranya refuses to start"

docs/mtls.md
```diff
@@ -0,0 +1,96 @@
...
+Paths to the root certs and device cert will be provided in the daemon config.
+These certs will be loaded into the QUIC syncer module when the daemon loads.
+All QUIC syncer traffic will be authenticated by checking root cert signatures.
+
+daemon_config.toml:
+```
+[sync.quic]
+...
+root_certs=<root certs directory>
+device_cert=<device cert directory>
```
chip-so: Since this specifies a directory and we are being open to things generated by external PKI, we should describe this loading behavior. Some things to consider:

- Are we only trying to load certain extensions or any plain file? OpenSSL doesn't specify extensions and it's kind of an ad-hoc standard. I've seen `.pem` and `.crt` and there's probably others.
- What input formats do we support? Is it just what OpenSSL can load? (I think this is sufficient)
- Does it recurse or just look for files in this directory?
- Does it follows symlinks?
- Does it refuse to load keys with dangerous (world-readable) permissions? Does it warn?
- Sometimes a `.pem` file can be a key. Do we support that?

And to be clear I don't think we need to be permissive here. It's probably fine to ask operators to rename a `.crt` to `.pem` or convert their DER certs to PEM (though I am not sure this is always possible).

docs/mtls.md
djohnson-spideroak: Should we allow "lax" mode, bypassing verification? Requiring certs like this is going to increase the difficulty of setting up quick deployments for development.

docs/mtls.md
gknopf-aranya: Added a NIST reference

docs/mtls.md
```diff
@@ -0,0 +1,96 @@
...
+- Verify that device cert is signed by one of the root certs when daemon loads, rather than failing later during TLS authentication
+- Cert revocation. Syncing with a revoked cert only leaks team metadata. Devices can be removed from an Aranya team without the need for revoking certs.
+- Cert rotation/renewal
+
+## Certificate Generation
+
+mTLS root and device certs are generated externally via a user's existing PKI infrastructure.
+Device certs are signed by one of the root certs or an intermediate CA cert using the PKI infrastructure.
+
+We recommend using P-256 ECDSA certs generated from a secret key of at least 256 bits.
```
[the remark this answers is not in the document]
gknopf-aranya: Added a NIST reference

docs/mtls.md
gknopf-aranya: For simplicity of the initial implementation, I'd like to assume `.pem` certs and no recursion/symlinks. Checking permissions of the certs/keys wouldn't hurt, although it seems like a system-level/integration/deployment requirement that doesn't impact the Aranya requirements.

docs/mtls.md
gknopf-aranya: Prototyping a certgen tool with Claude for comparison. It is slightly easier to use than openssl and guarantees P-256 ECDSA keys are used:
```
# Create CA
certgen ca --cert ca.pem --key ca.key --ca-name "My CA"

# Create signed cert
certgen signed --ca-cert ca.pem --ca-key ca.key \
  --cert server.pem --key server.key \
  --cn server --dns example.com --ip 192.168.1.10
```

openssl:
```
# Create CA
openssl ecparam -genkey -name prime256v1 -out ca.key
openssl req -x509 -new -nodes -key ca.key -sha256 -days 365 \
  -out ca.pem -subj "/CN=My CA"

# Create signed cert
openssl ecparam -genkey -name prime256v1 -out server.key
openssl req -new -key server.key -out server.csr -subj "/CN=server"
openssl x509 -req -in server.csr -CA ca.pem -CAkey ca.key -CAcreateserial \
  -out server.pem -days 365 \
  -extfile <(echo "subjectAltName=DNS:example.com,IP:192.168.1.10")
rm server.csr ca.srl
```

chip-so: Worth noting that these invocations do not create an encrypted private key, which is probably something you'll want to offer.

Did claude generate the openssl command-lines too? For some reason it creates a serial file (`-CAcreateserial`) and then deletes it. (If you want OpenSSL to just use a random serial number, it will do that if you omit `-CAcreateserial`)

gknopf-aranya: At least for the initial implementation, we're going to use plaintext keys with a recommendation to use an HSM, an encrypted filesystem, and strict file permissions.

djohnson-spideroak (approved):
I dont see any major issues, I think for a first pass just getting the mTLS in as specified here is good.

mooreoak (approved):

PR #164, opened 2026-01-12
Policy book updates

### LINKED ISSUE #145: policy-book: Update operators and built-ins to new math functions
We've removed `+` and `-` (binary and unary). Update the docs to reflect this.

### LINKED ISSUE #154: policy-book: document `dynamic` marker for effect fields
I noticed while looking at the pest grammar that effect fields can be annotated with `dynamic`.

There's a comment in some policy code used for testing that mentions it briefly

> // the "dynamic" keyword is an annotation used to indicate to the consumer
    // that this value is not based on static event data and may change.

cc: @apetkov-so @chip-so @jdygert-spok

chip-so:
Closes #145, closes #154

Also includes a request for clarification on finish block behavior.

### DIFF
--- a/policy-book/src/revision-history.md
+++ b/policy-book/src/revision-history.md
@@ -4,7 +4,6 @@ This lists major changes to the documentation over time.
 
 ## Revision B (January 2026)
 
-- Added documentation for the `dynamic` keyword in effect fields.
 - Removed math operators and added built-in math functions.
 
 ## Revision A (October 2025)


policy-book/src/reference/top-level/effects.md
```diff
@@ -21,4 +21,8 @@ Effects delivered through the application API report the ID and recall
 status of the Command that emitted them. Because commands may be
 reevaluated after syncing and merging, effects can be issued multiple
 times. An application might use the command ID and recall status to
-filter out duplicate effects.
\ No newline at end of file
+filter out duplicate effects.
+
+The `dynamic` keyword is an annotation used to indicate to the effect
+consumer that the value is not based on static data and may change in
+future evaluations. It has no effect on policy execution.
```
jdygert-spok: I don't really get the motivation for `dynamic`. It'd be nice to have some more explanation of how it is expected to be used. Also, if it will continue to have no effect on policy execution, I wonder if it should become a more general attribute.

```
#[dynamic]
b string
```

Then we could use that for other things intended for outside tooling as well, e.g.

```
#[rust("DeviceId")]
x id
```

so that the interface generator makes that `DeviceId` instead of `BaseId`.

chip-so: As I recall, it was something @mooreoak requested fairly early in the design, but we've never actually used it. I think we're just assuming all fields can change so it doesn't serve any practical purpose for how we're using it. I'd be fine with removing it entirely.

policy-book/src/SUMMARY.md
```diff
@@ -22,6 +22,7 @@
 - [Expressions](reference/expressions.md)
   - [Operators](reference/expressions/operators.md)
   - [Functions](reference/expressions/functions.md)
+    - [Math Functions](reference/expressions/functions/math.md)
```
gknopf-aranya: Arithmetic would cover the current functions. Although there may be an intent to add other types of math operations in the future.

policy-book/src/reference/expressions/functions/math.md
```diff
@@ -0,0 +1,25 @@
...
+
+# `sub(x int, y int) option[int]`
+
+`sub` subtracts `y` from `x` and returns an `optional[int]`. The
+optional is `Some(x - y)` if subtraction succeeds without underflow, and
+`None` if it does not.
+
+# `saturating_sub(x int, y int) int`
+
+`saturating_sub` subtracts `y` from `x` and returns an `int`. Instead of
```
gknopf-aranya:
```suggestion
`saturating_sub` subtracts `y` from `x` and returns `x - y` an `int`. Instead of
```

gknopf-aranya (dismissed):
lgtm, left a couple comments

policy-book/src/reference/expressions/functions/math.md
chip-so: Is "subtracts `y` from `x`" not clear enough? How about instead of "`x - y`", I put "the difference"?

chip-so: I don't think this is one of the places where repetition provides much, as the operation should be pretty well understood by anyone reading it.

But how about, "`saturating_sub` calculates `x - y`, returning the difference as an `int`"?

jdygert-spok (dismissed):

jdygert-spok (approved):

PR #166, opened 2026-01-22
testing: distributed scale testing

djohnson-spideroak:


docs/distributed-scale-testing-spec.md
```diff
@@ -0,0 +1,504 @@
+# Aranya Distributed Scale Testing Specification
+
+## Overview
+
+In order to support more realistic testing scenarios, this document describes a server that can be used to spawn daemons and control them from multiple machines. Each test server exposes a REST API that can be used to command the daemons based on device ID. An orchestrator can use the test server to start remote daemons and add them to a test scenario.
```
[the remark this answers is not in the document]
djohnson-spideroak: That would require a REST API per daemon, this allows us to control multiple daemons on a server with one API.

docs/distributed-scale-testing-spec.md
```diff
@@ -0,0 +1,504 @@
...
+    servers: Vec<TestServer>,
+    daemons: HashMap<DeviceId, TrackedDaemon>,
+    teams: HashMap<TeamId, TrackedTeam>,
+}
+```
+
+### Orchestrator API
+
+```rust
+impl TestOrchestrator {
```
gknopf-aranya: APIs can be defined in a Rust `trait` instead of an `impl`.

docs/distributed-scale-testing-spec.md
```diff
@@ -0,0 +1,504 @@
...
+```
+
+---
+
+## Error Codes
+
+| Code | HTTP | Description |
+|------|------|-------------|
+| DAEMON_NOT_FOUND | 404 | Daemon doesn't exist |
+| DAEMON_NOT_RUNNING | 409 | Daemon stopped |
```
gknopf-aranya: How do we know if a daemon isn't running? Do we wait for an API call to timeout?

docs/distributed-scale-testing-spec.md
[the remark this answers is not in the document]
gknopf-aranya: From what I can tell, the purpose of the test server appears to be bridging the gap between a remote orchestrator and a daemon's local IPC API. So the test server would likely run on the same machines that the daemons it controls are running on.

docs/distributed-scale-testing-spec.md
gknopf-aranya: While orchestrating daemons has some utility, it would be ideal if the test orchestrator could orchestrate both daemons and client applications to run full test scenarios.

docs/distributed-scale-testing-spec.md
```diff
@@ -0,0 +1,504 @@
...
+                                  │ HTTP/REST
+                    ┌─────────────┼─────────────┐
+                    │             │             │
+                    ▼             ▼             ▼
+           ┌──────────────┐ ┌──────────────┐ ┌──────────────┐
+           │ Test Server  │ │ Test Server  │ │ Test Server  │
+           │   (Node A)   │ │   (Node B)   │ │   (Node C)   │
+           └──────┬───────┘ └──────┬───────┘ └──────┬───────┘
+                  │                │                │
+        ┌─────────┼────────────┐  ...             ...
```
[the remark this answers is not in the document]
djohnson-spideroak: It uses the client library to talk to the daemons. Part of the reason to expose this as a test server that reverse-proxies to the daemons is to reduce the number of ports we need to grab. It also makes it much easier to manage, as without the test server the daemons would need to be manually started on remote machines.

docs/distributed-scale-testing-spec.md
djohnson-spideroak: The test server tracks which daemons are running. This code would be returned if you called `/stop` on a specific daemon. I think an IPC timeout should give a 500 error.

docs/distributed-scale-testing-spec.md
```diff
@@ -0,0 +1,504 @@
...
+[logging]
+level = "info"
+format = "pretty"
+```
+
+CLI: `aranya-test-server [-c config.toml] [-p port] [-w work_dir] [-v]`
+
+---
+
+## Sync Topologies
```
gknopf-aranya: I thought @benz-spideroak had a doc that describes different sync topologies and whether to use hello, poll, push, etc. sync methods. While this is a helpful example of how to setup a sync topology with the orchestrator, I don't think it's needed in the test orchestrator spec because the test orchestrator can be developed and meet requirements without this section.

docs/distributed-scale-testing-spec.md
```diff
@@ -0,0 +1,504 @@
...
+| Method | Endpoint | Description | Returns |
+|--------|----------|-------------|---------|
+| GET | `/health` | Health check | 200 OK |
+| GET | `/status` | Server status | server_id, version, daemon counts |
+
+### Daemon Lifecycle
+
+| Method | Endpoint | Description | Request | Returns |
+|--------|----------|-------------|---------|---------|
+| POST | `/daemons` | Start daemon | name, sync config, afc config | device_id, sync_addr, uds_path, pid |
```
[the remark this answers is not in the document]
djohnson-spideroak: This would be passing the config values (or maybe even just the actual config struct) to the endpoint. The only requirements to using this is that the test server is started and there exists a daemon executable with the test server. Everything else (like writing the configs based on REST API requests) is handled by the test server.

docs/distributed-scale-testing-spec.md
benz-spideroak: I'd like to hear more about what this is trying to accomplish. If it's just to prove that sync works in certain scenarios it seems like there would be simpler ways to do it. 

The simplest thing would be to spawn processes for the daemons. That wouldn't scale to multiple machines so you might be better off using Docker containers. You could then control the containers by writing instructions to a file or having the containers pull instructions from a queue in Redis or similar.

PR #168, opened 2026-01-27
167: optimize find_needed_segments

benz-spideroak:


docs/find-needed-segments-optimization.md
```diff
@@ -0,0 +1,119 @@
...
+    for each have_location H:
+        if is_ancestor(S, H):   // Expensive BFS
+            skip S (peer has it)
+    if not skipped:
+        add S to result
+```
+
+In the `slow_test_4` test case this produces ~62,000+ `is_ancestor()` calls in 7 seconds, each doing BFS with file I/O.
+
+## Key Insight
```
gknopf-aranya: A small example showing the number of steps for the original and new algorithm would be useful. It would explain the new algorithm in a way that's easier to understand and provide the reader with more intuition on why it is more efficient.

docs/multi-daemon-convergence-test.md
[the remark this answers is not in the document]
benz-spideroak: It shouldn't be here.

docs/find-needed-segments-optimization.md
```diff
@@ -0,0 +1,167 @@
...
+
+| Aspect      | Bound                              |
+|:------------|:-----------------------------------|
+| Memory      | O(CAP) — two queues in existing `TraversalBuffers` |
+| Per segment | O(CAP) for queue operations        |
+| Total       | O(n × CAP) where n = segments visited |
+
+Since CAP is a compile-time constant (512), this is effectively O(n). This replaces O(n × m × k) where k was itself a graph traversal.
+
+Each entry grows from 16 bytes (`Location`) to 24 bytes (`(Location, bool)` with alignment padding), so memory overhead vs the current `TraversalBuffers` is ~8 KB.
```
jdygert-spok: I think we could avoid increasing the size of the entries by keeping the vec partitioned so every entry to one side of some index is covered and the other side is uncovered. Operations should only take a single swap plus increment/decrement of that index.

PR #181, opened 2026-03-06
finalization: add BFT consensus-based finalization spec

### LINKED ISSUE #179: finalization: write spec for BFT consensus based finalization
Write a spec for finalization. Planning to use a BFT consensus algorithm.
Most promising Rust library so far seems to be `malachite` (well-maintained, stand-alone BFT algo):
https://github.com/circlefin/malachite

cc: @djohnson-spideroak, @mooreoak

gknopf-aranya:
## Summary

Specification for BFT consensus-based finalization using Tendermint via the malachite library.

### Finalization Policy (on-graph)
- Certified commands: new command type with author-independent command ID, `seal_certified`/`open_certified`/`verify_certified_quorum` FFIs
- Finalize command: carries `factdb_merkle_root`, committed via policy action at agreed-upon parent (action-at-parent)
- Finalizer set: established in Init, changed via two-phase UpdateFinalizerSet/Finalize flow
- Policy queries: `query_finalize_seq`, `query_finalizer_set`, `verify_finalization_proposal`

### BFT Consensus Protocol (off-graph)
- Tendermint-based via malachite library (with maturity risk analysis and build-vs-adopt tradeoffs)
- Two-phase voting (prevote/precommit) with detailed split-brain prevention rationale
- Certified signatures only at precommit (SIG-003) to prevent cross-round signature accumulation attacks
- Proposer rotation by sorted finalizer key ID
- Daemon relies on aranya-core for consensus message signing/verification (COMM-009)
- Malachite integration pseudocode with handler functions and consensus round loop

### Scheduling
- Delay-based (monotonic counter, no wall clock dependency, immune to NTP spoofing)
- Daemon restart triggers finalization unconditionally, leveraging quorum memory
- Command-count alternative discussed as future hybrid option

### Threat Model
- Attack vectors: malicious proposer, blocking, equivocation, MitM, replay, non-finalizer impersonation, slow clock, NTP spoofing, quorum loss, precommit signature safety, malformed envelopes
- Replay protection: consensus messages via Height+Round, certified signatures via parent_id in digest
- Quorum loss recovery analyzed (owner override unsafe, weighted voting ineffective)

### Key Design Decisions
- Certified commands required (author-signed commands cause graph forks under partition)
- Any finalizer can commit (not just proposer) for availability
- Consensus protocol signing delegated to aranya-core (consistent with sync architecture)
- Finalization scheduling does not rely on wall clocks or FactDB-persisted state

## Test plan
- [ ] Review spec for completeness and correctness
- [ ] Validate policy pseudocode against existing Aranya policy patterns
- [ ] Confirm malachite library integration approach with implementation team

Closes #179

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+| 2 | 2 | 0 |
+| 3 | 3 | 0 |
+| 4 | 3 | 1 |
+| 7 | 5 | 2 |
+| 10 | 7 | 3 |
+| 13 | 9 | 4 |
+
+### Changing the Finalizer Set
+
+The finalizer set can only be changed through a `Finalize` command that includes an optional `new_finalizer_set` field. This field, when present, contains the complete new set of finalizer public signing keys. The change takes effect at the next finalization sequence number.
```
gknopf-aranya: If a new finalizer set isn't specified, the current finalizer set remains unchanged.

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+
+Properties:
+
+- **Priority**: 0 (processed before all non-ancestor commands in the weave).
+- **Fields**:
+  - `seq` -- The finalization sequence number.
+  - `signatures` -- List of signatures from finalizers that agreed to this finalization. Each entry contains a signing key and signature. Only finalizers that participated are included.
+  - `new_finalizer_set` -- Optional. If present, the complete new set of finalizer public signing keys that takes effect at the next sequence number.
+- **Policy checks**:
+  - The `signatures` list contains at least `(n * 2 / 3) + 1` valid signatures from unique members of the current finalizer set.
```
gknopf-aranya: Use quorum instead of formula

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+  - No `FinalizeRecord` exists at this sequence number (prevents duplicates).
+  - The sequence number is sequential (previous seq must be finalized, or seq is 1).
+  - If `new_finalizer_set` is present, it contains at least 1 member.
```
gknopf-aranya: At least 4 members for fault tolerance. If this check fails, the original finalizer set remains unchanged.

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+
+- **Seq** -- The finalization sequence number.
+- **Round** -- The round number within this sequence number (increments on timeout).
+- **Finalization point** -- The graph command up to which the weave is finalized.
+- **Parent seq** -- The last finalized sequence number (0 if no prior finalization).
+
+**Prevote.** Each finalizer receives the proposal and independently verifies it by computing the weave from the same starting point to the proposed finalization point. If the finalizer can verify the proposal (it has all the commands and they produce a valid weave), it broadcasts a prevote for the proposal. If it cannot verify the proposal (due to missing commands, different graph state, or invalid proposal), the finalizer prevotes nil.
+
+Finalizers must prevote nil immediately for proposals that are obviously invalid -- for example, a sequence number that has already been finalized, or a proposal referencing an unknown parent. This allows the round to fail fast without waiting for the full timeout. If a quorum prevote nil, the round advances immediately to the next proposer.
```
gknopf-aranya: A potential optimization could be to send the latest graph head on a `nil` vote. This could give the next proposer more up-to-date info in case they have synced more commands since the last consensus round.

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+
+```policy
+command Finalize {
+    attributes {
+        priority: 0
+    }
+
+    fields {
+        seq int,
+        signatures list[struct FinalizerSignature],
```
gknopf-aranya: Does the policy support arrays like this?

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+            create FinalizeRecord[
+                seq: this.seq,
+            ]=>{}
+
+            // Update finalizer set if a new one was provided.
+            if this.new_finalizer_set is Some {
+                let new_set = unwrap this.new_finalizer_set
+
+                // Remove all current finalizers.
+                for f in get_finalizers() {
```
gknopf-aranya: Policy doesn't have iteration. May need an FFI for this.

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+}
+```
+
+The QUIC server reads the `MsgType` when accepting a stream and routes it to the appropriate protocol handler. Streams with an unrecognized `MsgType` are closed immediately.
+
+#### Finalizer Peer Configuration
+
+Finalizer network addresses are configured at runtime via the client API:
+
+- **`add_finalizer_peer(signing_key, address)`** -- Registers a finalizer peer's network address. The local finalizer establishes (or reuses) a QUIC connection to this address for consensus communication.
```
gknopf-aranya: `pub_signing_key_id`. Can use the ID of the signing key here. Should already know what the public key is based on the finalizer set.

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+- **`remove_finalizer_peer(signing_key)`** -- Removes a finalizer peer.
```
gknopf-aranya: We'll need a way to automatically remove network address configuration for finalizers that are no longer part of the finalizer set after a new finalization command is published.

docs/finalization.md
```diff
@@ -0,0 +1,483 @@
...
+The consensus protocol is implemented using the malachite library. The integration maps Aranya concepts to malachite abstractions:
+
+| Malachite Concept | Aranya Mapping |
+|---|---|
+| `Value` | Finalize command content (seq, finalization point) |
+| `ValueId` | Hash of the proposed Finalize command content |
+| `Height` | Finalization sequence number (seq) |
+| `Validator` | Finalizer device |
+| `ValidatorSet` | All devices in the finalizer set (equal voting power) |
+| `Address` | Device ID |
```
gknopf-aranya: Is the address supposed to be the unique identifier of the current finalizer? I think this should be the public signing key ID.

docs/finalization.md
```diff
@@ -0,0 +1,583 @@
...
+command Init {
+    fields {
+        // ... existing fields ...
+        finalizer1 optional bytes,
+        finalizer2 optional bytes,
+        finalizer3 optional bytes,
+        finalizer4 optional bytes,
+        finalizer5 optional bytes,
+        finalizer6 optional bytes,
+        finalizer7 optional bytes,
```
gknopf-aranya: Starting with a list of 7 finalizers seems overly verbose for an MVP. Perhaps we can start with 4 then increase to 7 once we nail down the rest of the implementation details.

docs/finalization.md
```diff
@@ -0,0 +1,691 @@
...
+
+In the future, finalizers may not need to be devices at all, since all that matters is a signing key. For MVP, finalizers are assumed to be devices.
+
+### Initialization
+
+The initial finalizer set is established in the team's `Init` command. The `Init` command includes 7 optional finalizer fields (`finalizer1` through `finalizer7`), each containing a public signing key ID. The caller specifies 1 to 7 finalizers. If none are specified, the team owner's public signing key ID is used as the sole initial finalizer.
+
+### Validator Set
+
+All devices in the finalizer set form the validator set. Each finalizer has equal voting power of 1. The maximum supported set size is 7.
```
gknopf-aranya: Why are we defining both a finalizer and a validator set? Aren't these the same thing? It's confusing to use multiple different terms for the same set. Maybe this is originating from the BFT algorithm terminology.

docs/finalization.md
```diff
@@ -0,0 +1,691 @@
...
+- **`verify_finalize_quorum(envelope)`** -- Reads the signatures from the multi-author envelope. Verifies each signature against the current finalizer set and the command content. Returns true if at least a quorum of valid, unique finalizer signatures are present.
+- **`validate_new_finalizer_set(f1..f7)`** -- Validates the new finalizer set fields. Checks that the specified count is between 1 and 7, all key IDs are unique, all correspond to valid signing key IDs, and the new set does not shrink below 4 if currently at 4+. Returns true if valid.
+- **`stage_pending_finalizer_set_update(command_id, f1..f7)`** -- Creates or replaces a `PendingFinalizerSetUpdate` fact with the command ID of the `UpdateFinalizerSet` command and the specified finalizer key IDs. If a pending update already exists, it is replaced.
```
gknopf-aranya: Why is this an FFI? Seems like something the policy can do by itself.

docs/finalization.md
```diff
@@ -0,0 +1,691 @@
...
+5. **Prevote.** Each finalizer verifies the proposal by computing the weave to the proposed finalization point:
+   - A, C: Can verify. Broadcast prevote for the proposal.
+   - B: Also prevotes for its own proposal.
+   - D: Missing some commands. Prevotes nil (includes its current head so the next proposer knows D's state).
+
+6. **Precommit.** A, B, and C each observe 3 prevotes for the proposal (quorum = `(4 * 2 / 3) + 1` = 3). Each broadcasts a precommit.
+
+7. **Decision.** All finalizers observe 3 precommits for the proposal. Agreement is reached.
+
+8. **Signature collection.** Each finalizer constructs the Finalize command content (seq=3, apply_finalizer_set_update=None), signs it, and shares the signature. Each collects at least 3 signatures.
```
gknopf-aranya: Isn't the finalization command supposed to include the finalization point so they all agree on finalizing the graph at the same point?

docs/finalization.md
```diff
@@ -0,0 +1,646 @@
...
+- **`stage_pending_finalizer_set_update(f1..f7)`** -- Creates or replaces the `PendingFinalizerSetUpdate` fact with the given finalizer key IDs.
+- **`maybe_apply_pending_finalizer_set_update()`** -- If a `PendingFinalizerSetUpdate` fact exists, deletes all current `Finalizer` facts, creates new `Finalizer` facts from the pending update's fields, and deletes the `PendingFinalizerSetUpdate` fact. No-op if no pending update exists.
+- **`verify_factdb_merkle_root(expected_root)`** -- Computes the FactDB Merkle root at the current evaluation point and returns true if it matches the expected root. Used by the Finalize command's policy block to verify that the proposer's claimed state matches the local state.
+- **`next_finalize_seq()`** -- Returns the next finalization sequence number (highest existing `FinalizeRecord` seq + 1). Derives the sequence number from the FactDB rather than requiring it in the command fields.
```
gknopf-aranya: You can derive this directly from the factdb without an FFI

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+- Commands on branches that conflict with the finalized weave are permanently recalled.
+- Devices can truncate graph data for finalized commands, retaining only the Finalize command as a compact proof of the finalized state.
+
+### FactDB Merkle Root Verification
+
+The Finalize command payload has two fields: `parent_id` (the command ID of the parent, which is effectively a Merkle root of the graph up to that point) and `factdb_merkle_root` (the FactDB Merkle root at that parent). In a correct implementation these are redundant -- the same `parent_id` should always imply the same FactDB state. `factdb_merkle_root` is included during initial development to detect any divergence between graph and FactDB state before it reaches production. Once unit tests confirm the invariant, it is a candidate for removal.
```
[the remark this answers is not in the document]
gknopf-aranya: Is there a plan to eventually support distributing the factdb to other nodes?

chip-so (commented):
Still going through this, but there's enough here I thought I should submit early.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+
+**Known limitation:** Because finalizer set changes are applied through the `Finalize` command, a set change cannot proceed if finalization itself is stalled.
+
+## Finalization Policy
+
+This section defines the on-graph commands, facts, and policy rules that enforce finalization. The policy is evaluated independently by every device -- it does not depend on or interact with the off-graph consensus protocol.
+
+### Multi-Author Commands
+
+The Finalize command uses multi-author authentication instead of single-author authentication. Each signature represents an author -- a finalizer that endorsed the command. This is necessary because finalization represents agreement by a quorum of finalizers, not the action of a single device.
```
mooreoak: I think I would say "certifier" vs "author" to make the distinction clear. Certified commands have no author. It might actually be interesting to make `merge` commands certified. With a requirement that a active device certified them. Right now an infinite number of merge commands could be created by anyone in the sync group.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+- **Fields**:
+  - `parent_id` -- The command ID of the parent of the Finalize command. Because each command ID is a hash of its content and its own parent's ID, `parent_id` is effectively a Merkle root of the entire graph up to that point -- it uniquely and cryptographically identifies the full chain of commands being finalized. All ancestors of this command become permanent.
+  - `factdb_merkle_root` -- The FactDB Merkle root at the parent of the Finalize command. In theory, the same `parent_id` should always produce identical FactDB state -- the two are redundant in a correct implementation. This field is included during initial development so that finalizers can detect graph/FactDB divergence before voting, surfacing any such bugs in pre-production. Once unit tests confirm that `parent_id` always implies a consistent FactDB state, this field is a candidate for removal as a production optimization (see [Future Work](#future-work)).
```
mooreoak: This is incorrect. 

We have the field so that we can a device can onboard form any finalized command instead of the init command. By including the merkle root it can  directly sync the factdb and verify it's correctness.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+- **Envelope**: Contains multiple `(signing_key_id, signature)` pairs from the finalizers that authored this command. Only finalizers that participated are included.
+- **Policy checks**:
+  - The envelope contains at least a quorum of valid signatures from unique members of the current finalizer set.
+  - The `factdb_merkle_root` matches the locally computed FactDB Merkle root at the parent of the Finalize command.
+  - The derived sequence number (`latest_finalize_record().seq + 1`) is sequential.
+- **Side effects**:
+  - Creates a `FinalizeRecord` at the next sequence number (derived from existing records).
+  - If a `PendingFinalizerSetUpdate` fact exists, applies it atomically (see [Changing the Finalizer Set](#changing-the-finalizer-set)). The Merkle root check guarantees all finalizers agree on whether a pending update exists.
+  - Emits an effect if the device's own finalizer set membership has changed (added or removed), so the daemon can immediately update its consensus participation state.
```
mooreoak: It should alway emit the whole set as it will change which devices the consensus need to be computed with.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+
+Branches do not finalize in parallel. Parallel finalization would produce Finalize commands that are not ancestors of each other, violating the chain guarantee. The graph must converge through merges before commands on separate branches can be finalized.
+
+### Post-Finalization
+
+Once a Finalize command is committed to the graph:
+
+- All commands that are ancestors of the Finalize command are permanently accepted. Their effects in the FactDB are irreversible -- the runtime enforces this through weave ordering rather than by rejecting new commands. The `Finalization` attribute ensures the Finalize command is always ordered before any siblings in the weave, and any new command must be appended to the Finalize command or one of its descendants -- never before it. This means new commands can only extend the graph forward from the finalized point, never insert before it.
+- Commands on branches that conflict with the finalized weave are permanently recalled.
+- Devices can truncate graph data for finalized commands, retaining only the Finalize command as a compact proof of the finalized state.
```
mooreoak: This is enabled but not strictly true with only this spec. I would not include it.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+        )
+
+        // Create initial FinalizeRecord so seq 1 has a predecessor.
+        finish {
+            create FinalizeRecord[seq: 0]=>{}
+        }
+    }
+}
+
+fact Finalizer[pub_sign_key bytes]=>{}
```
mooreoak: Why are you defining these facts and the bottom of these blocks? I think you should define all the facts and their relationships before the commands which manipulate them.

mooreoak (changes requested):
This is the first half of my review that covers the policy portions of this preposal.

This is overall good but misunderstands the point of the merkle root, and overcomplicates the needed new FFIs.

docs/finalization.md
```diff
@@ -0,0 +1,1105 @@
...
+
+### Triggering Finalization
+
+Finalization is triggered in three ways:
+
+1. **Periodic scheduling.** The `Init` command emits an effect that schedules the first finalization after a configured delay. Each successful `Finalize` command emits an effect scheduling the next one. Scheduling uses a delay (not wall clock time) -- the daemon tracks the delay locally. Only one finalization can be scheduled at a time; a shorter delay overwrites the current schedule.
+2. **Daemon restart.** When a finalizer daemon comes back online, it always attempts to trigger finalization. The daemon does not try to determine whether the last finalization was recent -- it proceeds unconditionally and relies on other finalizers to prevote nil if finalization was too recent. This avoids the daemon needing to persist scheduling state across restarts.
+3. **On-demand.** Any finalizer can initiate a finalization round on-demand by signaling other finalizers via the off-graph consensus protocol.
```
mooreoak: I believe on demand finalization must be due to an effect else it will be rejected as too soon.

docs/finalization.md
```diff
@@ -0,0 +1,1105 @@
...
+- **Parent** -- The graph command that the Finalize command will be appended to. This is not a payload field -- it is the position in the graph where the Finalize command is placed. All ancestors of the Finalize command become permanent.
+- **FactDB Merkle root** -- The FactDB Merkle root at the proposed parent. Agreement on the Merkle root implies agreement on the sequence number, finalizer set, and any pending updates -- all are derivable from the FactDB.
+
+**Sync and validate.** Each finalizer receives the proposal. If a finalizer does not have the proposed parent in its local graph, it syncs with the proposer to obtain it and all ancestor commands. Once the finalizer has the proposed parent, it validates the proposal by computing and comparing the FactDB Merkle root (see [Pre-Consensus Validation](#pre-consensus-validation)).
+
+**Prevote.** If validation passes, the finalizer broadcasts a prevote for the proposal to all other finalizers. A finalizer prevotes nil if: validation fails or the FactDB Merkle root does not match. Proposals with a stale or future height are handled by Malachite (dropped or buffered, respectively) before reaching validation.
```
mooreoak: Can it just send it's signature/certification at this time?

docs/finalization.md
```diff
@@ -0,0 +1,1105 @@
...
+
+Note: Standard Tendermint (as implemented by Malachite) requires a full quorum of nil prevotes to advance immediately. A smaller number of nil prevotes (e.g. `n - q + 1`, which would make proposal quorum mathematically impossible) does not trigger early advancement — the round waits for the timeout in that case. This is conservative but avoids giving a minority the ability to accelerate round advancement.
+
+**Decision.** When a quorum of precommits is observed for the same proposal, the consensus round reaches agreement. If precommit quorum is not reached (nil quorum or timeout), the consensus round number increments and a new proposer is selected. The process repeats from the proposal step with the new proposer.
+
+#### Phase 2: Signature Collection
+
+Once agreement is reached on the parent of the Finalize command, each finalizer deterministically constructs the Finalize command payload from the agreed-upon FactDB Merkle root. Because the Merkle root is deterministic given the parent, every finalizer produces the same payload and therefore the same command ID.
```
mooreoak: The proposer should have just sent the command to them.

docs/finalization.md
```diff
@@ -0,0 +1,1105 @@
...
+- **Truncation** -- Define a garbage collection strategy for finalized graph data. Requires Merkle roots (e.g. weave Merkle root) to retain compact proofs of truncated state.
+- **Light clients** -- Devices that verify Finalize commands without replaying the full weave.
+- **Larger finalizer sets** -- Support finalizer sets beyond the current maximum of 7. This requires policy language support for collection types or additional FFI work to handle more fields.
+- **Recursive types for finalizer set** -- If the policy language adds support for recursive types (e.g. `struct PubKeyList { key bytes, next option[PubKeyList] }`), the fixed 7-field finalizer set representation could be replaced with a variable-length linked list. This would also remove the need for the `FinalizerSet` singleton's fixed field slots and simplify the `UpdateFinalizerSet` and `PendingFinalizerSetUpdate` definitions.
```
mooreoak: This goes against some of the design principles of the policy langeage.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+- **Multiple authors.** Instead of a single `get_author()` check, the policy verifies that the envelope contains a quorum of valid signatures from the current finalizer set. Each signature is an author endorsing the command.
+- **New FFI functions.** Multi-author commands require new envelope FFI:
+  - `seal_multi_author(payload)` -- Seals a multi-author command. The command ID is computed from the payload as usual. The envelope is created without signatures; they are attached later.
+  - `open_multi_author(envelope)` -- Opens a multi-author envelope and returns the deserialized fields.
+  - `verify_finalize_quorum(envelope)` -- Reads the signatures from the envelope and verifies a quorum against the current finalizer set.
+
+### Finalize Command
+
+The `Finalize` command makes all of its ancestors permanent. It is the only graph command produced by finalization.
```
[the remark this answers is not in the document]
gknopf-aranya: Updating wording

mooreoak (changes requested):
I think the biggest gap in this spec is describing how we will get the certifications on the commands added to the graph. 

I also don't think we need to separate out the pre-commit and sign phase.

docs/finalization.md
gknopf-aranya: True, each finalizer needs to know the entire set of finalizer peers to send consensus messages to.

docs/finalization.md
```diff
@@ -0,0 +1,1105 @@
...
+- **During partitions**: Non-finalizer devices continue publishing and syncing. If fewer than a quorum of finalizers can communicate, finalization halts but graph operations are unaffected.
+- **After partition heals**: Devices sync and merge branches. Finalization resumes once a quorum can communicate. Finalizers that are missing the proposed parent sync with the proposer before voting -- this affects liveness, not safety.
+
+### Equivocation Detection and Vote Visibility
+
+**Equivocation.** Malachite detects conflicting votes from the same finalizer. This does not halt consensus; the protocol continues with honest finalizers.
+
+**Vote logging.** Each finalizer logs votes observed during each consensus round: prevotes for/nil, precommits for/nil, equivocation evidence, and non-responsive nodes.
+
+**Operator response.** Operators review vote logs to identify misbehaving finalizers. A device with the `UpdateFinalizerSet` permission can remove them via `UpdateFinalizerSet`.
```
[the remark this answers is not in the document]
gknopf-aranya: This seems to tie into observability goals. Identifying all the data to log would require some more design effort. We likely won't know all the information to include in a command like this until we have more familiarity with the malachite library.

docs/finalization.md
```diff
@@ -0,0 +1,1105 @@
...
+    Sync,
+    Consensus,
+}
+```
+
+The QUIC server reads the `MsgType` when accepting a stream and routes it to the appropriate protocol handler. Streams with an unrecognized `MsgType` are closed immediately.
+
+Finalizers open consensus streams only with other finalizers -- non-finalizer peers are unaffected and never see consensus traffic.
+
+#### Finalizer Peer Configuration
```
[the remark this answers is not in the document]
gknopf-aranya: I think we can. At one point, we had discussed wanting to support finalizers that are not devices. Since the MVP of finalization can assume that each finalizer is a device, we should be able to re-use sync peers for it.

docs/finalization.md
```diff
@@ -0,0 +1,1098 @@
...
+    new_finalizer6_pub_sign_key option[bytes],
+    new_finalizer7_pub_sign_key option[bytes],
+}
+```
+
+#### New FFIs
+
+FFI functions must be pure functions — they take inputs and return outputs without side effects. Each FFI below exists because the operation cannot be expressed in the policy language. All fact operations and effect emissions are performed in policy code.
+
+- **`validate_finalizer_keys(f1..f7)`** -- Takes up to 7 `option[bytes]` keys. Returns -1 if no keys are provided or if any non-None keys are duplicates; otherwise returns the count of non-None keys. The caller checks `n > 0` to reject invalid sets. *Required as FFI because the policy language lacks iteration — counting non-None optional fields and checking all 21 pairwise combinations cannot be expressed inline.* Used by `Init` and `UpdateFinalizerSet` commands. **TODO:** Confirm whether the policy language can propagate FFI errors directly. If so, this FFI should return an error instead of -1.
```
gknopf-aranya: @jdygert-spok, @chip-so is it possible for an FFI function to return an error or is it better to return -1 if this fails?

docs/finalization.md
gknopf-aranya: Yeah, it could schedule a finalization with zero delay

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+| **Non-finalizer impersonation** | A non-finalizer node sends consensus messages to influence voting | All consensus messages are signed by the sender's signing key and verified against the current finalizer set. Messages from non-finalizers are dropped. |
+
+## Finalizer Set
+
+The finalizer set is the group of devices authorized to participate in finalization consensus. Finalizer devices do not need to be members of the team -- they only need a signing key to participate in consensus and can sync graph commands like any other device. This allows dedicated finalization infrastructure that is separate from the team's member devices. The team creator (or owner, for later updates) must know the finalizer's public signing key at the time the finalizer is added to the set. The finalizer set can only be changed by devices with the `UpdateFinalizerSet` permission through a dedicated `UpdateFinalizerSet` command.
+
+In the future, finalizers may not need to be devices at all, since all that matters is a signing key. For MVP, finalizers are assumed to be devices.
+
+### Initialization
```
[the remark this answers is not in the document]
gknopf-aranya: The `malachite` library provides a way to access the size of the quorum via `ThresholdParam::min_expected`.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+
+**Sync and validate.** Each finalizer receives the proposal. If a finalizer does not have the proposed parent in its local graph, it syncs with the proposer to obtain it and all ancestor commands. Once the finalizer has the proposed parent, it validates the proposal by computing and comparing the FactDB Merkle root (see [Pre-Consensus Validation](#pre-consensus-validation)).
+
+**Prevote.** If validation passes, the finalizer broadcasts a prevote for the proposal to all other finalizers. A finalizer prevotes nil if: validation fails, the proposed parent does not advance beyond the last Finalize command, or the FactDB Merkle root does not match.
+
+If a quorum prevote nil, the round advances immediately to the next proposer.
+
+**Precommit.** Every finalizer independently observes the prevotes. When a finalizer observes a quorum of prevotes for the same proposal, it broadcasts a precommit for that proposal to all other finalizers. If a quorum of nil prevotes is observed, or the prevote timeout expires without quorum, the finalizer precommits nil. All finalizers participate in both voting stages.
```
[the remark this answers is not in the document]
gknopf-aranya: If we wait until consensus is reached before collecting signatures, then a quorum of signatures does imply consensus was reached via the BFT algorithm.

docs/finalization.md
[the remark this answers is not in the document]
gknopf-aranya: Can this be added to future work?

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+
+## Finalization Policy
+
+This section defines the on-graph commands, facts, and policy rules that enforce finalization. The policy is evaluated independently by every device -- it does not depend on or interact with the off-graph consensus protocol.
+
+### Multi-Author Commands
+
+The Finalize command uses multi-author authentication instead of single-author authentication. Each signature represents an author -- a finalizer that endorsed the command. This is necessary because finalization represents agreement by a quorum of finalizers, not the action of a single device.
+
+This requires a **new command type** in the `aranya-core` runtime. Existing commands assume a single author with a single signature. The multi-author command type extends the envelope to carry multiple signatures instead of one.
```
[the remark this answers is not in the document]
gknopf-aranya: Updating the spec with this info. Also starting to add details about the aranya-core types/methods needed to support the new command and open/seal operations.

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+
+Key properties of multi-author commands:
+
+- **Signatures live in the envelope, not the payload.** Single-author commands already store the signature in the envelope, separate from the payload. Multi-author commands follow the same pattern: the envelope carries up to 7 optional `(signing_key_id, signature)` pairs, matching the maximum finalizer set size. The payload contains only the command fields (parent_id, factdb_merkle_root). Because the command ID is computed from the payload, and signatures are in the envelope, the ID computation does not change. Different valid subsets of author signatures produce the same command ID, which is critical -- multiple finalizers may independently commit the same Finalize command with different signature subsets, and the policy must treat them as the same command.
```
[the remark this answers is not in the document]
gknopf-aranya: This has been resolved

docs/finalization.md
```diff
@@ -0,0 +1,1033 @@
...
+- **Multiple authors.** Instead of a single `get_author()` check, the policy verifies that the envelope contains a quorum of valid signatures from the current finalizer set. Each signature is an author endorsing the command.
+- **New FFI functions.** Multi-author commands require new envelope FFI:
+  - `seal_multi_author(payload)` -- Seals a multi-author command. The command ID is computed from the payload as usual. The envelope is created without signatures; they are attached later.
+  - `open_multi_author(envelope)` -- Opens a multi-author envelope and returns the deserialized fields.
+  - `verify_finalize_quorum(envelope)` -- Reads the signatures from the envelope and verifies a quorum against the current finalizer set.
```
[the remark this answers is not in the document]
gknopf-aranya: I've updated the spec to clarify that finalization commands should be committed as soon as a quorum of signatures is received. This means it is unlikely for a finalization command to have more than a quorum of signatures unless it receives multiple signatures at the same time (a race), causing them to be included in the finalization command before committing it. The reason for committing right away after receiving a quorum is that this reduces latency since there isn't a good way to know how long to wait for additional signatures and the quorum already provides the desired level of safety.

This implies that the policy will only need to verify a unique set of signatures as large as the quorum. Validating additional signatures that may be present on the command isn't a requirement. If an invalid signature is encountered, it should be skipped so signature validation continues to the next signature in the list. This ensures signature validation passes as long as there is a unique, valid set of signatures >= the quorum even if there may be invalid signatures in the list that were maliciously introduced.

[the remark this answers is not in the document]
gknopf-aranya: Planning to change this to `verify_certified_quorum()`

docs/finalization.md
gknopf-aranya: Looking into this more. We don't currently have a way to access a command directly after `seal()` via the policy interface. We interact with the policy via invoked actions and returned effects. It seems like this would require a new API.

docs/finalization.md
```diff
@@ -0,0 +1,1105 @@
...
+Finalization and normal graph operations are independent:
+
+- **During partitions**: Non-finalizer devices continue publishing and syncing. If fewer than a quorum of finalizers can communicate, finalization halts but graph operations are unaffected.
+- **After partition heals**: Devices sync and merge branches. Finalization resumes once a quorum can communicate. Finalizers that are missing the proposed parent sync with the proposer before voting -- this affects liveness, not safety.
+
+### Equivocation Detection and Vote Visibility
+
+**Equivocation.** Malachite detects conflicting votes from the same finalizer. This does not halt consensus; the protocol continues with honest finalizers.
+
+**Vote logging.** Each finalizer logs votes observed during each consensus round: prevotes for/nil, precommits for/nil, equivocation evidence, and non-responsive nodes.
```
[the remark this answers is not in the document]
gknopf-aranya: Added this to future work. Let me know if this is a requirement for MVP of the finalization feature.

docs/finalization.md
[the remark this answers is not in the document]
gknopf-aranya: Added this to future work. Let me know if this is a requirement for MVP of the finalization feature.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+└─────────────────────────────────────────────────────┘
+```
+
+Key architectural boundaries:
+
+- **Consensus manager** and **sync manager** orchestrate their respective protocols and deliver messages via the shared QUIC transport. They are daemon-layer components.
+- **Consensus protocol** and **sync protocol** are transport-agnostic crates in the `aranya-core` repository. Both produce and consume messages that the daemon delivers via the transport layer. Neither depends on QUIC directly.
+- **QUIC transport** is shared by both managers. The daemon multiplexes consensus and sync streams on the same QUIC connections.
+- **Finalization policy** is part of the daemon layer and depends on the aranya-core runtime.
+- **Finalization FFIs** are an optional runtime plugin for operations the policy language cannot express directly (certified envelopes, quorum verification).
```
gknopf-aranya: While it's optional for the application to enable the finalization FFI in general, if they want the finalization feature, the FFI must be enabled or the policy that implements finalization will not compile.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+
+In the future, finalizers may not need to be devices at all, since all that matters is a signing key. For MVP, finalizers are assumed to be devices.
+
+### Initialization
+
+The initial finalizer set MUST be established in the team's `Init` command. **[FSET-001]** The `Init` command MUST support 1 to 7 optional finalizer fields (`finalizer1` through `finalizer7`), each containing a public signing key. **[FSET-002]** The caller specifies 1 to 7 finalizers. If no finalizers are specified in `Init`, the team owner's public signing key MUST be used as the sole initial finalizer. **[FSET-003]** All specified finalizer keys MUST be unique and valid public signing keys. **[FSET-011]**
```
gknopf-aranya: These requirements redundantly specify the 1-7 finalizer requirement

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+| `q` | `⌊(n * 2) / 3⌋ + 1` | Quorum size -- the minimum number of finalizers required for a consensus decision. Ensures safety as long as at most `f` finalizers are Byzantine. |
+
+## Scope
+
+Finalization MUST apply only to persistent control plane commands on the DAG -- the commands that manage team membership, roles, labels, and AFC channel configuration. Ephemeral commands and AFC data plane traffic MUST NOT be subject to finalization. **[FIN-003]**
+
+## Design Goals
+
+1. **Safety** -- A Finalize command requires a quorum of valid signatures from the current finalizer set. Because any two quorums overlap, no two valid conflicting Finalize commands can exist in the graph. Any device can independently verify a Finalize command by checking the signatures in its envelope.
+2. **Availability** -- Finalization tolerates up to `f` faulty or offline finalizers. Non-finalizer devices continue operating normally (publishing commands, syncing) regardless of whether finalization is active.
```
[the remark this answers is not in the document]
gknopf-aranya: The connotation I interpreted from faulty was a software bug or defective hardware, not malicious or Byzantine. Maybe it is general enough, but being more specific ties this into the threat model a bit more clearly.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+
+The set can grow or shrink freely, but once a team has 4 or more finalizers, the set MUST NOT shrink below 4. **[FSET-008]** 4 is the smallest set size with Byzantine fault tolerance (f=1), so this prevents losing BFT safety once established.
+
+The owner determines the initial set in `Init` and controls which devices receive the `UpdateFinalizerSet` permission.
+
+Because set changes are only applied atomically by `Finalize` commands, the finalizer set is always globally consistent -- all devices that have processed the same `Finalize` commands agree on the same set.
+
+If a second `UpdateFinalizerSet` is published before the next finalization, the `PendingFinalizerSetUpdate` fact MUST be replaced with the new values. **[FSET-010]** The next finalization round picks up whatever pending update exists at that point.
+
+**Known limitation:** Because finalizer set changes are applied through the `Finalize` command, a set change cannot proceed if finalization itself is stalled.
```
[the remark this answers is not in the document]
gknopf-aranya: Since finalization consensus is essentially an extension of the team owners authority, I wonder if we need a way for the owner to unilaterally update the finalizer set and perform a finalization in the case where finalization is stalled. The problem with this is I'm not sure how we would have global consistency when a sole owner finalizing if there is also a finalizer set attempting to publish a finalization command at the same time.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+
+## Finalization Policy
+
+This section defines the on-graph commands, facts, and policy rules that enforce finalization. The policy is evaluated independently by every device -- it does not depend on or interact with the off-graph consensus protocol.
```
gknopf-aranya: The off-graph consensus protocol gates adding the finalization command to the graph. But if a quorum of signatures can be obtained another way, the policy doesn't care. It's the application/daemon level that enforces consensus, rather than something enforced by the policy. The policy only enforces a quorum of finalizer signatures.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+Compare with regular commands: `digest = H("SignPolicyCommand-v1", author_sign_id, command_name, parent_id, payload)` and `cmd_id = H("PolicyCommandId-v1", digest, signature)`. The different domain separators prevent cross-protocol attacks (a regular command signature cannot be reused as a certified command signature, and vice versa). The exclusion of `author_sign_id` and `signature` ensures all certifiers produce the same digest and the same command ID regardless of who signs or which signatures are included.
+
+**`CertifiedEnvelope` struct:**
+
+```rust
+struct CertifiedEnvelope {
+    parent_id: CmdId,
+    command_id: CmdId,
+    command_name: String,
+    payload: Vec<u8>,              // Serialized command fields
```
gknopf-aranya: Is the payload supposed to be inside the envelop or separate? Please check how this is done for regular commands.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+- A new action type MUST accept a pre-built `CertifiedEnvelope` and a parent command ID for commit. **[CMT-004]** The runtime:
+  1. Verifies the specified parent exists in the local graph (action-at-parent).
+  2. Skips the seal block — the envelope is already constructed by the daemon.
+  3. Runs the command's `open` block (`open_certified`) to verify the command ID and deserialize the payload.
+  4. Runs the command's `policy` block (quorum verification, Merkle root check, sequence check, pending update application).
+  5. If all checks pass, commits the command to the graph at the specified parent.
+
+### Finalize Command
+
+The set of commands that happen before a Finalize command is strictly the set of its ancestors. This makes all ancestors permanent. The Finalize command is the only graph command produced by finalization.
```
gknopf-aranya: This is the technically correct wording. Other places that refer to commands before finalization as "permanent" should be updated to this wording.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+3. Multiple finalizers committing the same Finalize produce the same command ID (see [Certified Commands](#certified-commands)). When synced to other nodes, the graph recognizes them as the same command (standard sync deduplication) — no weaving or policy evaluation occurs.
+
+
+### Action-at-Parent
+
+The Finalize command must be committed at a specific parent — the graph command that consensus agreed upon — rather than at the current head of the graph. This is necessary because unrelated commands may arrive between consensus completing and the Finalize command being committed. If the Finalize command were always appended to the head, the parent could change from what was agreed upon, invalidating the FactDB Merkle root.
```
gknopf-aranya: This sounds like a requirement but doesn't have a requirement number associated with it.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+        }
+    }
+}
+```
+
+#### New FFIs
+
+FFI functions must be pure functions — they take inputs and return outputs without side effects. Each FFI below exists because the operation cannot be expressed in the policy language. All fact operations and effect emissions are performed in policy code.
+
+- **`validate_finalizer_keys(f1..f7)`** -- Takes up to 7 `option[bytes]` keys. Returns the count of non-None keys on success, or an error if no keys are provided or if any non-None keys are duplicates. *Required as FFI because the policy language lacks iteration — counting non-None optional fields and checking all 21 pairwise combinations cannot be expressed inline.* Used by `Init` and `UpdateFinalizerSet` commands.
```
gknopf-aranya: non-None -> Some or configured/set keys

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+A finalization round produces a single Finalize command for a specific sequence number. It may span multiple consensus rounds if proposals fail. Each consensus round is a single propose-prevote-precommit cycle. Since signatures are attached to precommit messages, a finalizer commits the Finalize command as soon as it has collected a quorum of signatures — without waiting for the consensus protocol to report a decision (see [Signature Timing Rationale](#signature-timing-rationale)).
+
+#### Pre-Consensus Validation
+
+Before voting in consensus, each finalizer MUST independently validate the proposed FactDB Merkle root **[VAL-001]** by evaluating the `VerifyFinalizationProposal` ephemeral command (see [FactDB Merkle Root Verification](#factdb-merkle-root-verification)) at the proposed parent. The Finalize command does not exist yet at this point -- finalizers are validating the proposer's claimed state, not a command. This ensures that signatures certify verified state agreement, not just trust in the proposer.
+
+Finalizers that do not have the proposed parent MUST sync with the proposer to obtain it and all ancestor commands before validation. **[VAL-002]** Each finalizer then:
```
gknopf-aranya: If there are many commands missing, a single sync request (e.g. `sync_now()`) is not sufficient. A finalizer must continue syncing with the proposer until there are no more commands received and the graph head stabilizes.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+
+Consensus messages MUST be sent off-graph between finalizers. **[COMM-001]** The only on-graph command produced by finalization MUST be the `Finalize` command.
+
+#### Transport
+
+The consensus protocol MUST be transport-agnostic -- it produces and consumes messages that are delivered by an external transport layer, similarly to how the sync protocol is implemented. The daemon polls the consensus protocol for outgoing messages and delivers incoming messages to it.
+
+The daemon reuses the existing QUIC transport that it already uses for the sync protocol to deliver consensus messages between finalizers. Consensus and sync streams are multiplexed on the same QUIC connections, with each stream beginning with a protocol discriminant to route to the appropriate handler. This is a convenience of the current daemon implementation, not a requirement of the consensus protocol.
+
+Finalizers MUST only open consensus streams with other finalizers. **[COMM-002]** Non-finalizer peers MUST NOT see consensus traffic.
```
gknopf-aranya: This assumes all the finalizers have the latest finalization command which has updated to the latest finalizer set (which is not a valid assumption). It is more accurate to say that finalizers should only send consensus traffic to peers in the finalizers set. And nodes must drop consensus traffic received if they are not in the finalizers set.

docs/finalization.md
```diff
@@ -0,0 +1,837 @@
...
+- **`new_proposal`** -- Constructs a proposal containing the proposed parent and FactDB Merkle root.
+- **`new_prevote` / `new_precommit`** -- Constructs vote messages signed with the finalizer's signing key.
+
+### Library Selection
+
+Several Rust BFT consensus libraries were evaluated:
+
+| Library | Algorithm | Pros | Cons |
+|---|---|---|---|
+| [malachite](https://github.com/circlefin/malachite) | Tendermint | Standalone embeddable library with no runtime or networking opinions. Co-designed with TLA+ formal specs. Actively maintained (Informal Systems / Circle). Implements the same Tendermint algorithm used in production by 100+ Cosmos ecosystem blockchains (via CometBFT). | Tendermint's all-to-all voting has O(n^2) message complexity. Malachite itself is newer and does not have the same production track record as CometBFT. |
```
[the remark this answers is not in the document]
gknopf-aranya: If the risk is too great, then the tradeoff would be whether we should reimplement Tendermint ourselves, not whether a better library selection is available.

docs/finalization.md
```diff
@@ -0,0 +1,1133 @@
...
+                        let envelope = build_finalize_envelope(
+                            proposal.parent_id,
+                            proposal.merkle_root,
+                            accumulated_signatures,
+                        )
+                        commit_at_parent(envelope, proposal.parent_id)
+
+                        // Advance to next height. [CONS-011]
+                        next_height += 1
+                        accumulated_signatures.clear()
```
gknopf-aranya: This only works if there wasn't an error that caused execution not to make it this far. Better to drop it out of scope.

docs/finalization.md
```diff
@@ -0,0 +1,1166 @@
...
+                Precommit => {
+                    // Feed vote to Malachite for state management. [CONS-010]
+                    // Malachite validates height/round.
+                    consensus.receive_precommit(msg.precommit)
+
+                    // Accumulate certified signature. Precommit messages
+                    // always carry a certified signature (nil precommits
+                    // are a separate message type with no signature).
+                    let (key_id, sig) = msg.certified_signature
+                    accumulated_signatures.insert(key_id, sig)
```
gknopf-aranya:
```suggestion
                    accumulated_signatures.insert(msg.certified_signature)
```

docs/finalization.md
```diff
@@ -0,0 +1,1178 @@
...
+let next_height = query_finalize_seq() + 1
+let finalizer_set = query_finalizer_set()
+let my_finalizer_id = pub_key_bundle.sign_key.id()
+
+// Initialize the consensus protocol with the keystore. aranya-core
+// handles all signing/verification internally. [COMM-009]
+let consensus = ConsensusProtocol::new(keystore, finalizer_set, my_finalizer_id, threshold_params)
+
+// --- Handler: incoming message from peer ---
+fn handle_recv(bytes, consensus, finalizer_set, next_height) -> Option<CommitResult> {
```
gknopf-aranya: Define functions first before startup code and the main loop.

docs/finalization.md
```diff
@@ -0,0 +1,1185 @@
...
+  - The Finalize command MUST contain exactly one payload field: `factdb_merkle_root` (typed as `struct FactRoot`) **[FPOL-002]** -- the FactDB Merkle root at the parent of the Finalize command. The parent command ID — the finalization point in the graph — is part of the command envelope (as `parent_id` in the `CertifiedEnvelope`), not a payload field. It is included in the certified digest (`H("CertifiedPolicyCommand-v1", command_name, parent_id, payload)`) and committed via action-at-parent (see [Action-at-Parent](#action-at-parent)). Everything else is either implicit in the DAG position or derivable from the FactDB state that the Merkle root certifies (sequence number from `LatestFinalizeSeq`, finalizer set from the `FinalizerSet` fact, pending updates from `PendingFinalizerSetUpdate`). Finalizers independently compute this from their local FactDB and verify it matches before voting in consensus (see [Pre-Consensus Validation](#pre-consensus-validation)). The Merkle root also enables FactDB distribution: new members can receive the FactDB at a finalization point and verify it against the Merkle root without replaying the entire graph (see [Future Work](#future-work)).
```
chip-so: This single list item is... a lot. "typed as" makes it sound like it's clarifying the textual representation, not the field type. The middle bit about `parent_id` doesn't seem useful to clarify here. The latter bit describing usage of the merkle root is probably relevant, but is maybe more usefully elaborated on where the verification happens in the policy point below.

chip-so (commented):
Still working through this.

docs/finalization.md
```diff
@@ -0,0 +1,1185 @@
...
+
+### Certified Commands
+
+The Finalize command MUST use certified authentication instead of single-author authentication. **[CERT-001]** Each signature represents a certifier -- a finalizer that endorsed the command. Certified commands have no author. This is necessary because finalization represents agreement by a quorum of finalizers, not the action of a single device.
+
+This requires a **new command type** in the `aranya-core` runtime. Existing commands assume a single author with a single signature, and the command ID is computed over the serialized fields, command name, parent ID, author sign ID, and signature. The certified command type differs in two ways: it extends the envelope to carry multiple certifier signatures instead of one, and the command ID MUST be computed from the payload only — excluding signatures and signer identities. This exclusion is critical because different finalizers may independently commit the same Finalize command with different signature subsets, and all must produce the same command ID.
```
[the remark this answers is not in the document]
gknopf-aranya: Agree that single-author finalization commands (where the author ID is part of the command ID) will not work for this reason. If the author ID is not included in the command ID and the proposer is just responsible for committing the finalization command that should work, but has the availability drawbacks mentioned above.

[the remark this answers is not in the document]
gknopf-aranya: Summarizing where the discussion has landed so far:
- Add support for certified commands with different signature sets (with no author)
- Allow any finalizer to commit the same finalization command with a different signature set

docs/finalization.md
```diff
@@ -0,0 +1,1185 @@
...
+| Proposer | The finalizer selected by the BFT protocol's deterministic round-robin to propose a parent for the Finalize command for a given consensus round. The proposer may or may not be the initiator. |
+| Prevote | First-stage vote indicating a finalizer considers the proposal valid |
+| Precommit | Second-stage vote indicating a finalizer is ready to commit the proposal |
+| Byzantine finalizer | A finalizer that is malicious or arbitrarily faulty — it may crash, withhold messages, send conflicting votes, or otherwise deviate from the protocol. The BFT consensus protocol guarantees safety as long as the number of Byzantine finalizers does not exceed `f`. |
+| Fault tolerance (`f`) | The maximum number of Byzantine finalizers the protocol can tolerate. Derived as `⌊(n - 1) / 3⌋` (see [Formulas](#formulas)). |
```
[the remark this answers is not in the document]
gknopf-aranya: Yep, formulas should be captured in one spot

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+- **`action(command, parent: Some(command_id))`** — The command MUST be appended to the specified parent instead of the head. **[CMT-007]** The runtime MUST verify the specified parent exists in the local graph before committing; if it does not exist, the action MUST fail. **[CMT-008]**
+
+**FactDB evaluation context.** When using action-at-parent, the runtime MUST evaluate the command's policy against the FactDB state at the specified parent, not the FactDB state at the current head of the graph. **[CMT-010]** This is critical for finalization: the `verify_factdb_merkle_root` check and `FinalizerSet` lookup in the Finalize command's policy must operate on the FactDB at the agreed-upon parent, which may differ from the head if additional commands have been appended since consensus completed. If the runtime maintains a single FactDB reflecting the head, it MUST be able to reconstruct the FactDB state at the specified parent for action-at-parent evaluation. The runtime reconstructs the FactDB at a given point by replaying graph commands from a weave up to that point — snapshotting the FactDB at every command would require too much memory.
```
chip-so, line 256: I'm a little confused by this requirement, because I believe this is how the runtime already has to work in order to evaluate incoming commands whose parents are not the current head. "If the runtime maintains a single FactDB reflecting the head..." is immediately contradicted by the next sentence that states exactly how the FactDB works.

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+
+The Finalize command is the only command that uses the explicit parent mode. The daemon passes the agreed-upon parent from consensus when committing the Finalize command.
+
+Action-at-parent may create a sibling of the current head (increasing graph width), since other commands may have been appended to the head after the agreed-upon parent. This does not violate graph width assumptions in a meaningful way for finalization, because the `finalize: true` attribute ensures the Finalize command is always ordered first among siblings in the braid. The next command any device appends will be a descendant of the Finalize command (or a descendant of a later head), so graph width returns to normal immediately.
```
chip-so, line 260: Maybe this has been discussed elsewhere, but does this imply that the runtime creates a new head that merges the previous head and the action-at-parent Finalize command?

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+
+### FactDB Merkle Root Verification
+
+The `verify_factdb_merkle_root` FFI obtains the current FactDB Merkle root and compares it to the expected value. The runtime implements this FFI using the storage API, which computes the Merkle root incrementally and stores it on disk as part of the storage format.
+
+Before consensus, each finalizer validates the proposal using an ephemeral command that verifies the Merkle root without persisting anything to the graph:
```
chip-so, line 266: "without persisting anything to the graph" is implied by "ephemeral". (Maybe "ephemeral command" is a candidate to put in the Terminology section, but I feel like anyone reading this should already be familiar with the term)

chip-so, line 266: Also, does this need to be an ephemeral command and FFI? If it's ephemeral, it's not stored. It's not making a decision based on stored facts. Couldn't the daemon check the proposed merkle root at the given head by asking the storage impl directly? (I took the earlier "...which computes the Merkle root incrementally and stores it on disk as part of the storage format" to mean that this value is part of the storage implementation and can be queried for any command, but maybe I'm wrong.)

docs/finalization.md
```diff
@@ -0,0 +1,1185 @@
...
+
+A finalizer that has not yet synced the latest Finalize command may compute a different sequence number and therefore a different proposer. This does not break consensus — Malachite drops messages for a lower height and buffers messages for a higher height, so the stale finalizer's messages are harmless. The stale finalizer catches up through Malachite's built-in sync protocol: peers broadcast their current height, the behind node detects it is falling behind and fetches the missing decided values, then replays any buffered messages for the correct height. If the stale finalizer happens to be the designated proposer for a round, that round times out and rotation selects the next proposer — a liveness delay, not a safety issue. Safety is maintained as long as at most `f` finalizers are stale or faulty.
```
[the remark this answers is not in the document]
gknopf-aranya: The proposer does not need to sync before proposing (other finalizers sync with it after the proposal before prevoting). The spec already describes the process for how other finalizers should sync with the proposer. This process ensures aranya graphs are up-to-date during consensus.

The impact of the proposer not having an up-to-date graph before proposing is that an older command could be finalized. This is currently left up to the sync topology to determine how up-to-date the proposer is. Should we have the proposer sync with all the finalizers first?

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+fact PendingFinalizerSetUpdate[]=> {
+    num_finalizers int,
+    quorum_size int,
+    new_finalizer1_pub_sign_key option[bytes],
+    new_finalizer2_pub_sign_key option[bytes],
+    new_finalizer3_pub_sign_key option[bytes],
+    new_finalizer4_pub_sign_key option[bytes],
+    new_finalizer5_pub_sign_key option[bytes],
+    new_finalizer6_pub_sign_key option[bytes],
+    new_finalizer7_pub_sign_key option[bytes],
```
chip-so, line 347: Since this is structurally identical to `FinalizerSet`, maybe it makes sense to factor all of this out into a struct that's used by both?

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+        let next_seq = latest.seq + 1
+
+        // Check for a pending finalizer set update before the finish block.
+        // Agreement on the parent guarantees all finalizers agree on
+        // whether a pending update exists.
+        let pending = query PendingFinalizerSetUpdate[]
+
+        // Policy terminates after executing a finish block, so use
+        // branching to handle both cases in a single finish.
+        if pending is some {
```
chip-so, line 456: `some` needs to be capitalized.

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+                    f1_pub_sign_key: pending.new_finalizer1_pub_sign_key,
+                    f2_pub_sign_key: pending.new_finalizer2_pub_sign_key,
+                    f3_pub_sign_key: pending.new_finalizer3_pub_sign_key,
+                    f4_pub_sign_key: pending.new_finalizer4_pub_sign_key,
+                    f5_pub_sign_key: pending.new_finalizer5_pub_sign_key,
+                    f6_pub_sign_key: pending.new_finalizer6_pub_sign_key,
+                    f7_pub_sign_key: pending.new_finalizer7_pub_sign_key,
+                }
+                delete PendingFinalizerSetUpdate[]
+                emit FinalizerSetChanged {finalizer_set: finalizer_set}
```
chip-so, line 471: This is the _previous_ finalizer set. I'm not sure if this is what you wanted to do.

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+                // Schedule the next finalization round. [TRIG-002]
+                emit ScheduleFinalization {delay: configured_finalization_delay}
```
chip-so, line 473: `configured_finalization_delay` isn't defined and it doesn't sound like it should be the policy's responsibility anyway.

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+        )
+
+
+        // Stage the update. The next Finalize command will apply it
+        // automatically. If a pending update already exists, replace it.
+        let existing = query PendingFinalizerSetUpdate[]
+        if existing is some {
+            finish {
+                delete PendingFinalizerSetUpdate[]
+                create PendingFinalizerSetUpdate[]=>{
```
chip-so, line 533: Can't delete and create in the same finish block. Use `update`.

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+                }
+            }
+        }
+    }
+}
+```
+
+#### New FFIs
+
+FFI functions must be pure functions — they take inputs and return outputs without side effects. Each FFI below exists because the operation cannot be expressed in the policy language. All fact operations and effect emissions are performed in policy code.
```
chip-so, line 566: Feels more like instructions to Claude instead of useful information for the reader.

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+
+- **`validate_finalizer_keys(f1..f7)`** -- Takes up to 7 `option[bytes]` keys. Returns the count of provided keys on success, or an error if no keys are provided or if any provided keys are duplicates. *Required as FFI because the policy language lacks iteration — counting provided optional fields and checking all 21 pairwise combinations cannot be expressed inline.* Used by `Init` and `UpdateFinalizerSet` commands.
```
chip-so, line 568: This says it reports an error if no keys are provided, but it does not indicate how. Does this return a `result`? The prior code doesn't seem to suggest that.

docs/finalization.md
```diff
@@ -0,0 +1,1179 @@
...
+    }
+}
+```
+
+- **`query_finalize_seq`** -- Ephemeral action that emits `QueryLatestFinalizeSeqResult` with the last completed finalization sequence number. Used by the daemon on startup to derive the next Malachite height (`seq + 1`).
+- **`query_finalizer_set`** -- Ephemeral action that emits `QueryFinalizerSetResult` with the current finalizer set including public signing keys, count, and quorum size. Used by the daemon on startup to determine if this device is a finalizer and to build the Malachite validator set.
+
+#### Effects
+
+The `Finalize` command MUST emit a `FinalizerSetChanged` effect containing the current `FinalizerSet` using the policy `emit` statement. The daemon listens for this effect to update its consensus participation state and peer set without re-querying the FactDB.
```
chip-so, line 655: Feels like this should be further up, with the `Finalize` command definition.

chip-so (commented):
About half way through.

PR #182, opened 2026-03-16
spec: async onboarding

### LINKED ISSUE #186: spec: async onboarding

djohnson-spideroak:
This spec contains the process for onboarding a new device asynchronously. The goal is to allow a privileged device to onboard a new device by transferring an 11 word secret to the new device. Using this secret, the new device can publish their own keys to the graph and onboard themself. 

resolves #186


docs/async-onboarding.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Onboarding Sequence
+
+Actors:
+- Admin - the privileged device capable and authorized to initiate the asynchronous onboarding proceedure. 
+- Onboarding Server - the server that stores the onboarding bundle and validates the credentials provided for requests.
+- Onboarding Client - a client used to load a temporary keystore containing the self join key for the initial SelfJoinTeam action/command.
+- New Device - the device that is being onboarded to the team.
+
+```mermaid
```
gknopf-aranya: There is significant overlap between the information contained in the diagram, numbered list, and the requirements. Could be challenging to keep these in sync.

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
+# Async Onboarding
+
+Aranya currently requires synchronous exchange of information to onboard a new device. This specification provides a mechanism for devices to onboard themselves with a single exchange of information with a privileged device.
+
+The system uses an 11 word phrase to exchange entropy used to derive cryptographic material. The privileged device uses the derived key to encrypt an onboarding bundle. The encrypted onboarding bundle is then dropped in the onboarding server, and a one time key is added to the graph by the privileged device. The new device uses the 11 words (exchanged out of band) to derive the keys and information reiquired to fetch the onboarding bundle and decrypt it. The new device then uses the single-use onboarding key posted to graph to self join the team.
+
+
+## Architecture
+
+async-onboarding uses a standalone server to mediate asynchronous onboarding operations. This server is not a participant in the aranya team, but instead receives and distributes onboarding information according to a process that protects sensitive join information from the onboarding server.
```
gknopf-aranya: I'd mention that the server improves availability of the onboarding process.

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+sequenceDiagram
+    participant Admin
+    participant Graph
+    participant Server as Onboarding Server
+    participant Client as Onboarding Client
+
+    Note over Admin: Create one-time join keypair
+    Note over Admin: Create signed device certificate
+    Note over Admin: Generate 11-word phrase from CSPRNG
+    Note over Admin: Derive mailbox ID (128 bits)<br/>using HKDF-SHA-512
```
gknopf-aranya: How is the mailbox ID derived? Is it a hash of the 11 word phrase?

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+    Note over Admin: Derive symmetric encryption key<br/>using HKDF-SHA-512
```
gknopf-aranya: How is a symmetric key that is different from the mailbox ID derived from the same HKDF? Is there a salt or nonce included in this that results in a deterministic/unique derivation?

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+    Note over Admin: Derive authenticator<br/>using HKDF-SHA-512
+    Note over Admin: Encrypt cert + private key<br/>using the bundle key
+    Note over Admin: Encrypt one-time join keypair<br/>using the bundle key
+    Note over Admin: Encrypt pairing/syncing info<br/>using the bundle key
+    Note over Admin: Encrypt team ID<br/>using the bundle key
+
+    Admin->>Graph: Publish one-time onboarding public key<br/>(AllowSelfJoinTeam)
+    Admin->>Server: drop(mailbox ID, HMAC(authenticator, mailbox ID), ciphertext)
```
gknopf-aranya:
```suggestion
    Admin->>Server: store(mailbox ID, HMAC(authenticator, mailbox ID), ciphertext)
```

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+    Note over Client: Decrypt one-time join keypair
+    Note over Client: Decrypt pairing/syncing info
+    Note over Client: Decrypt team ID
+
+    Note over Client: Add team using decrypted team ID
+    Client->>Graph: Sync with peering point to get graph
+    Client->>Graph: SelfJoinTeam (using one-time join key)
+```
+
+1. Admin prepares onboarding process
```
gknopf-aranya: This numbered list looks almost identical to the sequence diagram I just reviewed. Do we need both? I think the sequence diagram is more expressive since it shows ordering better.

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+
+- Onboarding Client MUST be able to complete both the store and fetch proceedure
+
+**Onboarding Client store requirements**
+
+- Onboarding Client MUST be able to generate the 11 words for the admin
+- Onboarding Client MUST create a one time symmetric join key
+- Onboarding Client MUST create a signed device certificate
+- Onboarding Client MUST use a CSPRNG to create the 11 word phrase
+- Onboarding Client MUST use diceware to generate the entropy for the 11 word phrase.
```
gknopf-aranya: The requirements indicate using a CSPRNG and diceware to generate the 11 word phrase. Which one is generating the phrase?

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+
+1. Admin prepares onboarding process
+    1. Admin create one time join key (asymmetric key)
+    2. Admin create signed device certificate
+    3. Admin create onboarding bundle
+        1. Admin creates 11 word phrase from CSPRNG
+        2. Admin derives mailbox ID (128bits) using HKDF-SHA-512 and static salt
+        3. Admin derives symmetric encryption key (AES-256-GCM) for onboarding bundle using HKDF-SHA-512
+        4. Admin derives authenticator that the new user will use to authenticate to the onboarding server using HKDF-SHA-512
+        5. Admin encrypts certificate + private key using the bundle key
```
gknopf-aranya: Which key is used for encryption here? The private key or the symmetric key? Which key is being encrypted? The bundle key hasn't been defined yet so I'm not sure if that is the symmetric key, join key, or something else.

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+
+**Admin Requirements**
+
+- Admin MUST send 11 word phrase to new device out of band
+
+**Onboarding Client**
+
+- Onboarding Client MUST be able to complete both the store and fetch proceedure
+
+**Onboarding Client store requirements**
```
gknopf-aranya: I think we should add a requirement that `AES-256-GCM` use a `CSPRNG` to generate a nonce for each encryption. The encryption algorithm is not secure without this.

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+    3. Admin create onboarding bundle
+        1. Admin creates 11 word phrase from CSPRNG
+        2. Admin derives mailbox ID (128bits) using HKDF-SHA-512 and static salt
+        3. Admin derives symmetric encryption key (AES-256-GCM) for onboarding bundle using HKDF-SHA-512
+        4. Admin derives authenticator that the new user will use to authenticate to the onboarding server using HKDF-SHA-512
+        5. Admin encrypts certificate + private key using the bundle key
+        6. Admin encrypts one time join keypair using the bundle key
+        7. Admin encrypts pairing/syncing info using the bundle key
+        8. Admin encrypts team ID using the bundle key
+    4. Admin publishes one-time onboarding public key to graph (AllowSelfJoinTeam)
```
mooreoak: This includes all the fields that can be used when joining other than the keyid.

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+- Onboarding Server MUST validate the authenticator by calculating HMAC-SHA-512(authenticator, mailboxID) and comparing it to the value received from Admin when handing requests on the `fetch` endpoint
+- Onboarding Server MUST store the mailbox ID, HMAC of authenticator, and ciphertext.
+- Onboarding Server MUST expose a `store` endpoint that accepts a mailbox ID, the authenticator hash, and ciphertext
+- Onboarding Server MUST expose a `fetch` endpoint that accepts a mailbox ID and the authenticator.
+
+
+
+## Algorithms used
+
+- HKDF-SHA-512 for KDF
```
mooreoak: use sha-256 for these

docs/async-onboarding.md
```diff
@@ -0,0 +1,272 @@
...
+- HMAC-SHA-512 for HMAC
+- AES-256-GCM for symmetric encryption
+- Ed25519 for signing
```
mooreoak: use FIPS 204 (ML-DSA) for signing (from RustCrypto)

mooreoak (changes requested):
This is looking good. Needs to be fleshed out more to fill in missing details.

docs/async-onboarding.md
```diff
@@ -38,66 +41,71 @@ Actors:
...
+        2. Admin derives mailbox ID (128bits) using HKDF over the entropy and static info field
+        3. Admin derives symmetric encryption key for onboarding bundle using HKDF over the entropy and a different static info field. This is the "bundle key"
+        4. Admin derives authenticator that the new user will use to authenticate to the onboarding server using HKDF over the entropy and a static info field
+	5. Admin encrypts the onboarding bundle
+		1. Admin encrypts certificate + private key using the bundle key
+		2. Admin encrypts one time join keypair using the bundle key
+		3. Admin encrypts pairing/syncing info using the bundle key
+		4. Admin encrypts team ID using the bundle key
+    4. Admin publishes the public portion of the join key to the graph (AllowSelfJoinTeam) along with the values that will be associated with the new device like rank, role, etc.
+    5. Admin sends onboarding bundle to onboarding server, with mailbox ID, encrypted payload, CipherSuite ID, and HMAC of authenticator against mailbox ID
```
mooreoak: This is a second use of the term "onboarding bundle". The firs was about the plain text fields that were encrypted and this seems to be the cypher text + other values. We should pick different names for these to avoid confustion.

mooreoak (dismissed):

docs/async-onboarding.md
```diff
@@ -0,0 +1,287 @@
...
+
+The onboarding server then stores this data for use with the `fetch` endpoint.
+
+
+### `fetch`
+
+The `fetch` endpoint is used by new devices to fetch the encrypted onboarding bundle. This endpoint does not require authentication via PKI, and instead authenticates requests based on the provided authenticator. Fetch takes two arguments:
+
+1. The mailbox ID
+2. The authenticator
```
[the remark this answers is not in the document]
gknopf-aranya: Summarizing what I've learned about this so we can resolve it:
- Since the mailbox ID is used as an index to look up the onboarding bundle, it is intrinsically susceptible to timing attacks
- The authenticator is validated with a constant time equality check which is resistant to timing attacks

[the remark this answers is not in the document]
gknopf-aranya: I spent some time digging into the HMAC implementation to confirm that it requires equality checks to use a constant time comparison and a `==` comparison will not compile.

docs/async-onboarding.md
```diff
@@ -0,0 +1,287 @@
...
+2. The HMAC of the authenticator and mailbox
+3. The ciphertext of the onboarding bundle
+4. The CipherSuite ID
+
+The onboarding server then stores this data for use with the `fetch` endpoint.
+
+
+### `fetch`
+
+The `fetch` endpoint is used by new devices to fetch the encrypted onboarding bundle. This endpoint does not require authentication via PKI, and instead authenticates requests based on the provided authenticator. Fetch takes two arguments:
```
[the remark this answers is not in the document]
gknopf-aranya: Marking this as resolved since the timeout/TTL on store ensures data is eventually cleaned up on the server.

docs/async-onboarding.md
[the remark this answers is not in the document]
gknopf-aranya: The team discussed not doing this since it shifts time from the implementation to the spec. I think this should be fine. My concern is that describing the crypto less precisely could mean it's more difficult to ensure the crypto is correct before we start on the implementation.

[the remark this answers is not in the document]
mooreoak: The whole point of the cypher sweat and our crypto API is to safely abstract away the primitives used under the hood.

docs/async-onboarding.md
```diff
@@ -38,131 +134,97 @@ Actors:
...
+
+### Definitions
+
+**Onboarding Bundle** — the encrypted payload stored on the onboarding server, containing:
+- Device certificate + private key (signed by root CA)
+- Join key (one-time-use asymmetric keypair)
+- Pairing/syncing information
+- Team ID
+
+**Derived Keys** — all derived from the 11-word phrase using HKDF with distinct static info fields:
```
mooreoak, line 147: It would be good to list the static strings to put in to the info here.

docs/async-onboarding.md
```diff
@@ -2,33 +2,129 @@
...
+#### `fetch`
 
 The `fetch` endpoint is used by new devices to fetch the encrypted onboarding bundle. This endpoint does not require authentication via PKI, and instead authenticates requests based on the provided authenticator. Fetch takes two arguments:
 
 1. The mailbox ID
 2. The authenticator
 
+#### Onboarding Server Requirements
+
+SRV-002 Onboarding Server MUST authenticate users using a certificate when handling requests on the `store` endpoint.
```
mooreoak: I think it would be good to say "SRV-002 Onboarding Server MUST authenticate users using a mTLS client certificates when handling requests on the `store` endpoint.

docs/async-onboarding.md
```diff
@@ -38,131 +134,97 @@ Actors:
...
+    1. Admin creates the one time use "join key" (asymmetric key)
+    2. Device Certificate
+	    - Admin creates the certificate for the new device and signs it with the root CA if external PKI was not selected
+	    - If external PKI is in use, the application will use the certificate passed to the onboarding client for the device
+    3. Admin creates the "onboarding bundle" ciphertext
+        1. Admin creates 11 word phrase by encoding 128bits from CSPRNG using diceware
+        2. Admin derives mailbox ID (128bits) using HKDF over the entropy and static info field
+        3. Admin derives symmetric encryption key for onboarding bundle using HKDF over the entropy and a different static info field. This is the "bundle key"
+        4. Admin derives authenticator that the new user will use to authenticate to the onboarding server using HKDF over the entropy and a static info field
+	5. Admin encrypts the onboarding bundle containing
```
mooreoak, line 193: It would be good to say which CypherSweat API is used to do this.

docs/async-onboarding.md
```diff
@@ -264,7 +328,7 @@ function open_envelope_selfjoin(sealed_envelope struct Envelope) bytes {
 
 ## Diceware
 
-Diceware is used to generate a secure passphrase using a wordlist. The wordlist SHOULD be the EEF Large wordlist. Generating 11 words should provide 128 bits.
+Diceware is used to generate a secure passphrase using a wordlist. DCW-001 The wordlist SHOULD be the EEF Large wordlist. The diceware encoding will use 11 words to encode 128bits of information.
```
mooreoak, line 331: We might want to put the hash of the word list here as that is something that can be tested against.

mooreoak (dismissed):
I added a few comments. I don't see them as blocking but would improve the spec.

docs/async-onboarding.md
```diff
@@ -0,0 +1,336 @@
...
+#### `store`
+
+The `store` endpoint is authenticated by validating that the certificate presented matches a list of expected certificates provided at startup, and that it is signed by a specific root authority. Requests against this endpoint require authentication via PKI. Store takes the following arguments:
+
+1. The mailbox ID
+2. The HMAC of the authenticator and mailbox
+3. The ciphertext of the onboarding bundle
+4. An expiration time after which the onboarding bundle will no longer be available for fetching
+
+The onboarding server then stores this data for use with the `fetch` endpoint. SRV-001 After the expiration time has elapsed, the server MUST delete the stored bundle and reject any fetch requests for that mailbox ID.
```
chip-so, line 28: Should the server validate that the expiration time is not extremely far in the future?

docs/async-onboarding.md
```diff
@@ -0,0 +1,336 @@
...
+async-onboarding uses a standalone server to mediate asynchronous onboarding operations and provide greater availability during the onboarding process. This server is not a participant in the aranya team, but instead receives and distributes onboarding information according to a process that protects sensitive join information from the onboarding server. 
+
+The server provides two endpoints: `store` and `fetch`. These endpoints correspond to the privileged device storing an encrypted onboarding bundle, and the new device fetching the onboarding bundle. 
+
+#### `store`
+
+The `store` endpoint is authenticated by validating that the certificate presented matches a list of expected certificates provided at startup, and that it is signed by a specific root authority. Requests against this endpoint require authentication via PKI. Store takes the following arguments:
+
+1. The mailbox ID
+2. The HMAC of the authenticator and mailbox
```
chip-so, line 24: It is a little confusing that the "authenticator" isn't properly defined until much later. A brief description ("a one-time key derived from the 11 word phrase") and a link to the proper definition would be nice.

docs/async-onboarding.md
```diff
@@ -0,0 +1,336 @@
...
+The onboarding client provides an easy way to interact with the onboarding server. It provides the admin the ability to easily handle the required cryptography and store the onboarding bundle on the server. It also allows the new device to fetch the onboarding bundle from the server.
+
+
+#### Onboarding Client Requirements
+
+CLT-001 Onboarding Client MUST be able to complete both the store and fetch procedure.
+
+#### Onboarding Client Store Requirements
+
+STR-001 Onboarding Client MUST be able to generate the 11 words for the admin.
```
chip-so, line 68: I assume the 11 word requirement is related to the size of the wordlist, but there's no requirement about the size of the wordlist. I suppose it's implied by STR-009, but I think the math for why 11 words is required should at least be explained somewhere.

docs/async-onboarding.md
```diff
@@ -0,0 +1,336 @@
...
+
+STR-002 Onboarding Client MUST create a one time symmetric join key.
+
+STR-003 Onboarding Client MUST create a device certificate signed by the root CA if custom PKI is not in use.
+
+STR-004 Onboarding Client MUST use the provided input certificate if custom PKI is enabled.
+
+STR-005 Onboarding Client MUST optionally accept a certificate to use if custom PKI is enabled.
```
chip-so: I don't understand the difference between STR-004 and STR-005. Is one talking about a client cert and the other a root cert?

docs/async-onboarding.md
```diff
@@ -0,0 +1,336 @@
...
+
+STR-006 Onboarding Client MUST use a CSPRNG to create the 11 word phrase.
+
+STR-007 Onboarding Client MUST use at least 128 bits of entropy to generate the phrase.
+
+STR-008 Onboarding Client MUST validate 11 words exist in the phrase.
+
+STR-009 Onboarding Client MUST use the EFF large wordlist.
+
+STR-010 Onboarding Client MUST use diceware to generate the entropy for the 11 word phrase.
```
chip-so, line 84: Maybe make "diceware" a link to the explanatory section below.

docs/async-onboarding.md
```diff
@@ -0,0 +1,336 @@
...
+
+STR-011 Onboarding Client MUST derive a mailbox ID using HKDF from the 11 words.
```
chip-so, line 86: Does this mean HKDF on the number represented by the words, on UTF-8 data of the words, or something else? It's unclear.

docs/async-onboarding.md
```diff
@@ -0,0 +1,336 @@
...
+
+
+
+```
+
+
+
+## Diceware
+
+Diceware is used to generate a secure passphrase using a wordlist. DCW-001 The wordlist SHOULD be the EEF Large wordlist. The diceware encoding will use 11 words to encode 128bits of information.
```
chip-so, line 331: This doesn't mean [actually rolling physical dice](https://en.wikipedia.org/wiki/Diceware), does it? I'm not sure "diceware" is clear enough as an algorithm description, so this could probably use a reference or some elaboration.

chip-so (commented):
I think the keystore switcheroo is a little hacky, but I understand it would be difficult to rearchitect things to do it more properly. :/

docs/async-onboarding.md
```diff
@@ -0,0 +1,287 @@
+# Async Onboarding
+
+Aranya currently requires synchronous exchange of information to onboard a new device. This specification provides a mechanism for devices to onboard themselves with a single exchange of information with a privileged device.
```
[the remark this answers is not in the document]
gknopf-aranya: The timing attack mitigation is an example of an attack that was mitigated by the design that isn't obvious just from reading the spec.

PR #184, opened 2026-03-17
Draft of the plan to stabilize aranya-core

ssweetney-spok:


docs/core-stabilization-plan.md
```diff
@@ -0,0 +1,228 @@
...
+
+- **Timeline**: ???
+
+## Release Strategy
+
+Stabilization will proceed in **two phases**:
+
+| Phase                   | Scope                                                                                      | Goal                                                                                           |
+| ----------------------- | ------------------------------------------------------------------------------------------ | ---------------------------------------------------------------------------------------------- |
+| **Phase 1 (Immediate)** | Runtime, sync, storage, crypto engine, FFI modules, policy ifgen, Aranya IDs               | Stabilize the APIs users touch for the runtime + sync workflow to minimize compilation errors. |
```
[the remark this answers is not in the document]
ssweetney-spok: Potentially. I think we can stabilize the interface of the language without stabilizing the language though. There are breaking changes to the language we can make that won't have an affect on the interlace (and ifgen tooling). We should decide what we want to do here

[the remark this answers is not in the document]
gknopf-aranya: Yeah, it seems possible. Caveat is that it is risky to stabilize a layer in the architecture that depends on another layer that is not stabilized. This seems like a risk worth mentioning.

docs/core-stabilization-plan.md
```diff
@@ -0,0 +1,228 @@
...
+### 2. Sync Protocol APIs (`aranya-runtime`)
+
+- **Hide the wire format**: Add deserialization functions/methods so users do not need to depend on `postcard` to deserialize sync messages. This removes `postcard` from the user-facing API surface.
+
+- **State machine refactor**: Proposed but **still under discussion**. The idea is to refactor sync types to use a state machine to drive syncing.
+- **Pull out Sync APIs**: We may be able to pull the current set of APIs for syncing into a separate crate and release them consider those stable. The refactor mentioned above would be a breaking change that will be in a v2 (or the next major version of that crate).
+  
+**Open questions**:
+
+- [x] Will future sync optimizations change the sync APIs? (Ask Ben.) Example: `SyncResponder::find_needed_segments` was modified to take `TraversalBuffers`. ^dda3f8
```
gknopf-aranya: Whether sync protocol APIs will change doesn't seem relevant here. The purpose of the change is to allow internal sync protocol request/response/message types and wire format to be hidden behind the protocol API. So these kind of sync protocol changes won't block stabilization of the sync protocol API (though they could cause merge conflicts if both are in-progress at the same time).

docs/core-stabilization-plan.md
```diff
@@ -0,0 +1,228 @@
...
+	- Answered: Yes.
+
+---
+## Phase 2: Follow-up Work
+
+These items are **not** part of the initial stabilization release. They will be addressed in a subsequent release once designs and implementations have settled.
+
+### Policy Language (`aranya-policy-lang`, `aranya-policy-compiler`, `aranya-policy-vm`)
+
+The policy language has several planned **breaking** changes that must land before stabilization:
```
gknopf-aranya: The are some new FFIs anticipated when finalization is added.

docs/core-stabilization-plan.md
ssweetney-spok: Oh, you're right. I thought the function I mentioned here was public for some reason

ssweetney-spok:
These notes have been turned into an internal doc

PR #195, opened 2026-04-08
Close #187: Document result type in policy-book

### LINKED ISSUE #187: policy-book: Document result type
Now that we have the result type, we should add it to the policy book.

apetkov-so:


policy-book/src/reference/types.md
```diff
@@ -1,4 +1,64 @@
...
+    if n > 0 {
+        return Ok(n)
+    } else {
+        return Err("invalid n")
+    }
+}
+```
+
+### Coercion to Never
+If only one variant is specified, the missing type is inferred as `Never`. E.g. `Ok(42)` becomes `result[int, Never]`, and `Err("fail")` becomes `result[Never, string]`.
```
chip-so: This is more of a type checking implementation detail than something the programmer needs to know. We don't talk about "never" types at all in the book. I'd remove this section (and put it in a Policy Langonomicon if we had one).

policy-book/src/reference/types.md
```diff
@@ -1,4 +1,64 @@
 # Types
 
 Aranya Policy Language supports several basic and structured types.
-Notably, it does not support any collection types.
\ No newline at end of file
+Notably, it does not support any collection types.
+
+## Result type
```
jdygert-spok: The other types are in separate sections in their own files.

policy-book/src/reference/types.md
```diff
@@ -1,4 +1,64 @@
...
+    Ok(color) => {}
+    Err(msg) => {}
+}
+```
+
+We can also use expressions:
+```policy
+let x = match r {
+    Ok(n) => n
+    Err(e) => return Err(e)
```
chip-so: Expressions were used in the previous example (though not to any effect). This would be clearer if the previous example used only concrete cases, to show the difference.

policy-book/src/reference/types.md
```diff
@@ -1,4 +1,64 @@
...
+  ```
+  Ok(5) | Ok(6) // okay
+  ```
+  ```
+  Ok(5) | Ok(n) // invalid - binding (n) matches all values, so the literal is redundant
+  ```
+
+### Nesting
+Results can be nested:
+`result[result[int, string], enum Err]`. This is limited to one level.
```
chip-so: Is it? I don't recall any limits in the code.

policy-book/src/reference/types.md
jdygert-spok: Not sure we need to explicitly specify this. If we want to mention the `never` type then we probably want to give `never` its own section, although it should generally be invisible to the user. Maybe we just mention how result literals can be coerced to any matching type?

policy-book/src/reference/types.md
```diff
@@ -1,4 +1,64 @@
 # Types
 
 Aranya Policy Language supports several basic and structured types.
-Notably, it does not support any collection types.
\ No newline at end of file
+Notably, it does not support any collection types.
+
+## Result type
+A type that captures the outcome of an operation, and the resulting value or error: `result[OkType, ErrType]`.
```
chip-so: I'd expand on the literal syntax of `Ok()` and `Err()` here. It's used down in the Matching section but it's not really explained.

policy-book/src/reference/types.md
jdygert-spok: There's a limit?

policy-book/src/reference/types.md
jdygert-spok, line 1: Add railroad diagram like the other types?

chip-so:
Dygert and I did almost the exact same review at the exact same time. 😆

policy-book/src/reference/types.md
jdygert-spok: I think this is about match statement vs match expression, not about binding, but it could be more clear.

policy-book/src/reference/types.md
apetkov-so: My bad. We don't restrict result nesting, like we do option. I guess we'll want to add that.

apetkov-so: Yeah, I was thinking of options, but we didn't do it for result.

apetkov-so:
> Dygert and I did almost the exact same review at the exact same time. 😆

Great minds think alike, right? :)

policy-book/src/reference/types.md
apetkov-so: Moved

policy-book/src/reference/types/result.md
chip-so, line 1: So in order for this to show up in the book, it has to be added to the table of contents in `SUMMARY.md`.

policy-book/src/reference/types/result.md
```diff
@@ -0,0 +1,58 @@
...
+
+```policy
+function get_value(n int) result[int, string] {
+    if n > 0 {
+        return Ok(n)
+    } else {
+        return Err("invalid n")
+    }
+}
+```
```
djohnson-spideroak: I think we need like one or two more sentences, especially stating clearly what each type in the result type is used for. "A type that captures the outcome of an operation, and the resulting value or error." is also unclear.

policy-book/src/reference/types.md
chip-so: I think this still needs to be addressed.

chip-so (approved):

djohnson-spideroak (approved):

PR #196, opened 2026-04-14
Add policy lang `or`/`recall` spec

### LINKED ISSUE #194: policy-lang: Write `or` and `recall` spec
In some discussion, we decided that the solution to `check_unwrap` is to replace it with an `or` operator and a `recall` statement/expression. Write the spec for how these work.

cc: @mooreoak @jdygert-spok @djohnson-spideroak @apetkov-so

chip-so:


docs/policy-lang-or-recall.md
```diff
@@ -0,0 +1,129 @@
...
+
+    policy {
+        let user = get_user(this.user_id) or recall invalid_user(this.user_id)
+
+        if user.privilege_level < ADMIN_LEVEL {
+            recall insufficient_privileges(user)
+        }
+    }
+
+    recall invalid_user(user_id id) {
```
apetkov-so, line 104: What's the syntax for calling the unnamed recall block? I assume we're keeping that for compatibility.

chip-so: Yeah, I wanted to ask if we were doing that. In that case I think it would just be a bare `recall`. But for congruency (and maybe dumb Pest reasons) it should be `recall default`?

docs/policy-lang-or-recall.md
```diff
@@ -0,0 +1,129 @@
...
+function get_capability(object_id id) struct Capability {
+    let cap_opt = query Capability[object_id: object_id]
+    return cap_opt or default_capability(object_id)
+}
+```
+
+There are two subexpressions on either side of `or`. The left
+subexpression must be an `optional` type. If the left subexpression
+evaluates to `Some(v)`, it is unwrapped and `v` is the value of the
+expression. If it is `None`, then the value of the expression is the
```
apetkov-so, line 59: > If it is `None`, then the value of the expression is the value of the right subexpression.

What if the left expression evaluates to Never? E.g. let `x = return foo() or 5`.  We should probably disallow that.

chip-so: Yeah, it should be strictly `optional`, not `never`.

jdygert-spok: Yeah. We won't need to explicitly disallow it, we'll just match on the type to extract the inner type, like how we currently do in `lower_unwrap`.

docs/policy-lang-or-recall.md
jdygert-spok: 1. `recall`
2. `recall()`
3. `recall default`
4. `recall default()`

1 is the nicest IMO but is ambiguous in the current grammar. 2 looks a bit funny but I think is the simplest to implement. 3 and 4 maybe make default a keyword or at least maybe make it so you can't use it as a name in the scope of where you can reference the default recall?

jdygert-spok:
LGTM other than open question of default recall.

docs/policy-lang-or-recall.md
apetkov-so: I'm not a fan of 3 and 4 because the recall block isn't actually named "default". Well, the mangled name would be something like "Foo_recall_default", but that's not visible to the policy writer.

chip-so: I like 1 best too. Failing that, I'd go for 3 and make "default" an explicit keyword. That is, in the grammar it would be something like `"recall" ~ ("default" | function_call)`.

apetkov-so: With #3, we still can't pass args. The whole unnamed-block situation is awkward, both in terms of implementation (we have to either use a magic "default" name, or use Optional), and in terms of ergonomics. The only reason to keep it is compatibility.

But since we're introducing breaking changes anyway (removing check and check_unwrap, adding a new recall statement) we may as well require all recall blocks to be named. Thoughts?
+@mooreoak

chip-so: I thought default recall didn't have any args. How does the syntax for an unnamed block with args work?

I'm fine with breaking it. I think recall has been used very sparingly so far, if at all.

jdygert-spok: I didn't think we were removing check, and I thought we were keeping the default block for `check false` to call.

mooreoak: Ya I think all blocks being named is fine.

mooreoak: I also think that @jdygert-spok suggestion that for the interim we keep check which calls the default recall block and that is the way to reach it.

chip-so: > I didn't think we were removing check, and I thought we were keeping the default block for `check false` to call.

I thought we were changing `check` to require `else recall` on the end in policy context. I suppose that should probably be part of this spec too.

mooreoak (dismissed):

docs/policy-lang-or-recall.md
apetkov-so: > I also think that [@jdygert-spok](https://github.com/jdygert-spok) suggestion that for the interim we keep check which calls the default recall block and that is the way to reach it.

But that means we keep the unnamed recall. If that's the case, how about we make them all unnamed, but requiring unique signatures? Similar to catch blocks.

```
recall {}
recall(e enum Error) {}
```

These are invoked like so:
```
recall
recall(Error::whatever)
```

Or combined with check:
```
check x or recall(Error::whatever)
check x or recall // calls the no-arg block
check x // 🤔 not sure about this
```

chip-so: So recall blocks are identified purely by the argument type signature? I guess that works, but I don't find it very clear.

apetkov-so: Yeah, the idea is that the arg value (especially if it's an enum) provides enough context about the reason for the recall.

apetkov-so (approved):

docs/policy-lang-or-recall.md
```diff
@@ -0,0 +1,170 @@
+---
+layout: page
+title: "Policy Lang: 'or' and 'recall'"
+permalink: "/policy-or-recall/"
+---
+
+## The "never" type and control flow escape
```
chip-so, line 7: So I've kind of put this here to provoke some discussion. The behavior of "never" is kind of wild as we've implemented it, and maybe some specification (probably outside this one) should rein that in?

docs/policy-lang-or-recall.md
```diff
@@ -0,0 +1,170 @@
...
+
+## `check ... else ...`
+
+The expression in `check` is now followed by an `else` keyword and a
+terminal expression. Specifically, that means something that has the
+"never" type.
+
+```policy
+check authorized else recall unauthorized()
+check valid_user(user_id) else return Err(Error::InvalidUser)
```
chip-so, line 130: While reviewing the compiler I realized the other thing that produces "never" is the optional literal `None`. So technically this would be valid:

```
check foo else None
```

And I don't know what to do with that. Seems like there needs to be a distinction between "never" and "terminal".

jdygert-spok: `None` has type `option[never]`.

chip-so: Oooh, okay, right. So then could you do `check foo else None or return true`? Which is not a problem, I guess. It's just weird looking.

apetkov-so: That would compile, and it would be equivalent to `check foo else return true`.

chip-so: Yeah. Okay. No problem, then. 👍

docs/policy-lang-or-recall.md
jdygert-spok: Wrote some about a possible solution here: #198.

jdygert-spok: Alternatively, we implement an unreachable code lint. We could just only consider `return/recall` in that lint I guess, and not consider code after stubbed ffi or `todo` to be unreachable.

jdygert-spok (approved):

docs/policy-lang-or-recall.md
mooreoak: It feels like we need the type tree `any -> (never, value -> ...)` so we can have places that need to be `never`, `value` or `both`. (I did read @jdygert-spok 's issue)

docs/policy-lang-or-recall.md
```diff
@@ -0,0 +1,170 @@
...
+The `or` statement should provide a safe alternative to `unwrap` and `check_unwrap`, so they are removed from the language.
+
+### Migrating error handling
+
+Unwrap panicking is still possible by explicitly invoking `todo()`,
+which makes it more clear that the error is not being handled.
+
+```policy
+// let x = unwrap y
+let x = y or todo()
```
mooreoak, line 151: Do we have a lint or `--no-todo` compile flag so you can exclude `todo()` at compile time? If not we should create a ticket for that.

mooreoak (approved):
Some comments but no blockers.

docs/policy-lang-or-recall.md
jdygert-spok: Yes, the `debug` flag permits `todo` and turns on `debug_assert`. We also `tracing::warn!` regardless when compiling a `todo`.

docs/policy-lang-or-recall.md
chip-so: I like the further type specification and `any`. I think that's a good enough direction to satisfy me for this PR.

PR #199, opened 2026-04-20
improve skiplist

benz-spideroak:
- Rework the skip list design in docs/skiplist.md from randomized skip targets to deterministic exponentially-spaced targets (N/2, 3N/4, 7N/8, ...), producing ~log2(N/10) entries per segment.
- Add sparse construction: only ~1 in 10 segments builds a skip list; others rely on a nearby ancestor's list within MIN_SKIP_GAP (10) steps.                                                                       
- Specify merge-segment handling: walks start from the LCA instead of the parent, and the LCA is always recorded as a skip entry.                                                                                  
- Describe the O(log n) backwards walk used to collect skip targets by reusing intermediate segments' skip lists.                                                                                                     
- Remove the old random-skip description and the locate search pseudocode (superseded by the new construction rules).                                                                                       
- Add a "Skip List Optimization" section to docs/find-needed-segments-optimization.md that uses the skip list to jump from head to highest_have_max_cut + SEGMENT_BUFFER_MAX before the main traversal, reducing total cost across sync rounds from O(n^2) to O(n log n).

### DIFF
--- a/docs/find-needed-segments-optimization.md
+++ b/docs/find-needed-segments-optimization.md
@@ -165,3 +165,47 @@ Each entry grows from 16 bytes (`Location`) to 24 bytes (`(Location, bool)` with
 **Segment max_cut ranges**: The `Segment` trait exposes `shortest_max_cut()` (first command) and `longest_max_cut()` (last command). Heads entries carry `longest_max_cut` (via prior Locations pointing to segment heads) and are popped highest first. Pending entries carry `shortest_max_cut` (via `first_location()`). The flush condition compares pending `shortest_max_cut` against the current head's `longest_max_cut`, ensuring correctness when ranges overlap.
 
 **Result ordering**: The result may need a final sort to ensure causal order (parents before children).
+
+## Skip List Optimization
+
+The single-pass traversal described above is O(n) per sync round where
+n is the number of segments between the head and the peer's known
+commands. When syncing a large graph across many rounds (each
+transferring SEGMENT_BUFFER_MAX segments), the total cost is O(n^2)
+because each round re-walks from the head.
+
+### Approach
+
+Before starting the main traversal, use skip lists to jump from the
+graph head to approximately `highest_have + SEGMENT_BUFFER_MAX`. This
+skips over the large uncovered region that won't fit in the current
+batch anyway (only SEGMENT_BUFFER_MAX segments are collected per
+round).
+
+The jump works as follows:
+
+1. Compute the target: `highest_have_max_cut + SEGMENT_BUFFER_MAX`,
+   where `highest_have` is the max cut of the peer's most recent known
+   command.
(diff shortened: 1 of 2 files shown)


docs/skiplist.md
```diff
@@ -74,45 +70,65 @@ Segment {
...
-the segments parent will be checked. If a merge command is reached both
-parent segments will be added to the list of heads. The algorithm will then be
-repeated with all of the added heads.
+## Skip List Construction
 
-This will result in quickly skipping deeply into the graph. When the search gets
-close and can no longer skip it'll fall back to checking segments and if a
-segment is reached whose low_max_cut is less than the commands max cut, then the
-search will back up and continue from a previous head.
+### Skip Targets
```
mooreoak, line 104: I think this section should explain why the traditional probabilistic approach will not work. (ref https://en.wikipedia.org/wiki/Skip_list)

docs/find-needed-segments-optimization.md
```diff
@@ -165,3 +165,47 @@ Each entry grows from 16 bytes (`Location`) to 24 bytes (`(Location, bool)` with
 **Segment max_cut ranges**: The `Segment` trait exposes `shortest_max_cut()` (first command) and `longest_max_cut()` (last command). Heads entries carry `longest_max_cut` (via prior Locations pointing to segment heads) and are popped highest first. Pending entries carry `shortest_max_cut` (via `first_location()`). The flush condition compares pending `shortest_max_cut` against the current head's `longest_max_cut`, ensuring correctness when ranges overlap.
 
 **Result ordering**: The result may need a final sort to ensure causal order (parents before children).
+
+## Skip List Optimization
+
+The single-pass traversal described above is O(n) per sync round where
+n is the number of segments between the head and the peer's known
+commands. When syncing a large graph across many rounds (each
+transferring SEGMENT_BUFFER_MAX segments), the total cost is O(n^2)
```
mooreoak, line 174: Did we not ever update the implementation to allow more then one sync response message? I thought we had. If we have why are we limited by `SEGMENT_BUFFER_MAX`?

mooreoak (changes requested):

docs/skiplist.md
benz-spideroak: Added

docs/find-needed-segments-optimization.md
benz-spideroak: I don't think so. But I don't think that would affect this. Even with multiple response messages we'd still need to search the graph n/100 times to find the commands to send for each message. We can't walk the graph a single time without keeping the entire graph in memory.

mooreoak: I don't fully understand this but I don't think I need to right now.

mooreoak (approved):

djohnson-spideroak (approved):

PR #201, opened 2026-05-15
Lazy merges spec

benz-spideroak:
Spec for a multi-head storage model that defers merge authoring to local action time and rebuilds a fact cache lazily at consumer events, dropping multi-peer convergence from O(P*N^2) to O(P*N).

### DIFF
--- a/docs/lazy-merges.md
+++ b/docs/lazy-merges.md
@@ -105,6 +105,12 @@ update_head_set(cmd):
     head_set.retain(|h| !is_ancestor(h, cmd))
 ```
 
+### Merge ingestion
+
+A merge command carries only its parent addresses on the wire (`CommandMeta` in `sync/wire.rs`); the fact index is not transmitted, and a receiver has no way to verify a peer's fact index short of redoing the braid. So sync ingestion of a peer-authored merge braids the two parents locally and populates the merge segment's `FactIndex` from the result. This is the existing `add_merge` path in `Transaction::add_commands` (`crates/aranya-runtime/src/client/transaction.rs`) and is unchanged.
+
+In the lazy regime peers don't author merges during sync, so the merges on the wire are only those a peer wrote at commit time to collapse for a local action. Compared to eager — where every sync batch generates a peer-local merge that other peers then ingest — the merge volume drops sharply, which is what keeps the per-ingest braid cost affordable.
+
 ### Cache read and rebuild
 
 Every cache read goes through `read_cache`, which checks freshness and rebuilds if stale. Query, session, and author callers do not call a separate `ensure_cache_fresh`; they read the cache and the rebuild happens implicitly. Queries pass a `NullSink`; author and the public `refresh_cache` API thread the caller's sink so braid-fold effects flow out during rebuild.
@@ -138,7 +144,7 @@ Per-event costs:
 | Event | Cost |
 |:---|:---|
 | Sync ingest one command | O(1) head-set update plus per-segment write |
-| Sync ingest a merge command from a peer (both parents heads) | O(1) head-set update; no braid work |
+| Sync ingest a merge command from a peer | O(1) head-set update plus one braid over the two parents to populate the merge's fact index (see [Merge ingestion](#merge-ingestion)) |
 | Query, fresh cache | O(1) |
 | Query, stale cache | O(N + n) rebuild where N = head set size, n = commands since LCA |
 | Local author, fresh cache | (N-1) pairwise braids plus authored command write |
--- a/docs/lazy-merges.md
+++ b/docs/lazy-merges.md
@@ -37,9 +37,9 @@ Local authoring rebuilds the cache if stale, collapses the head set pairwise int
 
 ### Storage model
 
-The persisted graph state changes from `head: Location` to `heads: HeadSet`, where `HeadSet` is a small bounded vector of locations. The single-head case is a one-element head set.
+The persisted graph state changes from `head: Location` to `heads: HeadSet`, where `HeadSet` is a bounded vector of locations. The cap matches the braid's max-cut limit (256 concurrent authors); sync ingestion that would push the head set past capacity returns an error rather than forcing an inline merge — the same posture the braid takes today. The single-head case is a one-element head set.
 
-`Writer::head() -> Location` becomes `Writer::heads() -> HeadSet`. `Writer::commit(head: Location)` becomes `Writer::commit(heads: HeadSet, fact_cache: FactCacheRef)`.
+`Writer::head() -> Location` becomes `Writer::heads() -> HeadSet`. `Writer::commit(head: Location)` becomes `Writer::commit(heads: HeadSet, fact_cache: FactCacheRef)`; the fact cache is a separate persisted record (see [Fact Cache](#fact-cache)), and commit writes both atomically.
 
 ## Transaction Lifecycle
 
@@ -64,7 +64,9 @@ Init creates the graph and seeds `self.heads` with the init command directly; it
 
 ## Fact Cache
 
-The fact cache is a persisted record holding the merged fact perspective for some head set, separate from any merge segment. It contains:
+Today the merged perspective lives on the latest merge segment's fact index — every commit writes a merge, and that merge's `FactIndex` *is* the cache. Lazy merging removes the merge from the sync path, so the perspective has nowhere to live unless we give it a dedicated home.
+
+The fact cache is that home: a persisted record holding the merged fact perspective for some head set, separate from any merge segment. It contains:
 
 - A reference to the cached `FactIndex`.
 - The head set fingerprint at cache build time.
@@ -99,21 +101,8 @@ ingest(commands):
     // fact cache is left as-is; the next cache read rebuilds if stale
 
 update_head_set(cmd):
-    match cmd.parent():
-        Single(parent):
-            if parent in head_set:
-                replace parent with cmd in head_set
-            else:
-                add cmd to head_set
-                minimize head_set (remove ancestors)
-        Merge(left, right):
-            if {left, right} are both in head_set:
-                replace {left, right} with cmd in head_set
-            else:
-                add cmd to head_set
-                minimize head_set
-        None:
-            // init command; head set starts as {cmd}
+    head_set.insert(cmd)
+    head_set.retain(|h| !is_ancestor(h, cmd))
 ```
 
 ### Cache read and rebuild
--- a/docs/lazy-merges.md
+++ b/docs/lazy-merges.md
@@ -8,134 +8,54 @@ permalink: "/lazy-merges/"
 
 ## Overview
 
-Lazy merging defers writing merge commands until a local action requires a single parent. A graph carries a set of concurrent heads; sync extends the set without authoring merges. The merged fact perspective is held in a **fact cache** that is rebuilt lazily: sync updates the head set without folding policy, and the cache is rebuilt only when a query, session, or local author needs it. Sync bursts that aren't followed by a consumer event collapse into a single rebuild. Peers that have received the same authored commands have equal head sets regardless of arrival order, so cross-peer convergence completes in the rounds needed to propagate authored commands alone — no merge-of-merges cascade.
+Lazy merging changes two things:
+
+1. **Storage holds a head set and a fact cache** instead of a single head. The persisted `head: Location` becomes `heads: HeadSet`. A separate `FactCache` record holds the merged perspective for that head set so queries and sessions can read multi-head state at O(1).
+2. **`Transaction::commit` no longer collapses heads.** It writes `self.heads` to storage as-is, with a rebuilt fact cache. The pairwise-merge collapse moves to `action()`, where a new command actually needs a single parent.
+
+Everything else stays the same. `Transaction::add_commands` already accumulates a multi-head `self.heads` across multiple calls; the lifecycle API is unchanged. Orchestrators that hold a transaction open across multiple sync exchanges before committing get the most benefit, but the existing per-sync commit pattern still works.
 
 ## Problem
 
-`Transaction::commit` collapses every divergent head into a single graph head before returning, even in sync-only transactions. Each merge runs the braid algorithm over the divergent region between the existing graph head and the new commands.
-
-When P peers converge on N total commands, each peer ingests roughly N commands in batches from the other peers. Every batch triggers a braid whose region grows toward N as convergence progresses, and every braid writes a merge segment that subsequent syncs must walk through. Total work is O(P·N²).
-
-Profile of `generate_graph_hello_sync.test` (9 clients, 450 commands, no sync during generation, full convergence at end): ~2.5 million segment reads in the sync phase, close to the 9·450² = 1.82M lower bound.
-
-A related issue: peers that ingest the same authored commands in different orders produce structurally distinct merge commands. These merges then propagate, cascading into further merges before peers converge on the same state.
-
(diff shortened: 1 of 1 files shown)


docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+
+Sync ingestion updates the head set and lets the cache go stale. The cache is rebuilt on demand when a query, session, or author event needs it.
+
+Local authoring rebuilds the cache if stale, collapses the head set pairwise into a single head, writes the merge segments with their own fact indexes, and writes the authored command on top. The cache seeds the final merge's fact index.
+
+## Head Set
+
+### Storage model
+
+The persisted graph state changes from `head: Location` to `heads: HeadSet`, where `HeadSet` is a small bounded vector of locations. The single-head case is a one-element head set.
```
chip-so: What happens when `HeadSet` overflows? Does that also force a merge?

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+    // fact cache is left as-is; the next cache read rebuilds if stale
+
+update_head_set(cmd):
+    match cmd.parent():
+        Single(parent):
+            if parent in head_set:
+                replace parent with cmd in head_set
+            else:
+                add cmd to head_set
+                minimize head_set (remove ancestors)
```
chip-so: I think I can imagine how this minimize procedure would work, but maybe it's worth being explicit and describing it too?

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+layout: page
+title: Lazy Merges
+permalink: "/lazy-merges/"
+---
+
+# Lazy Merges
+
+## Overview
+
+Lazy merging defers writing merge commands until a local action requires a single parent. A graph carries a set of concurrent heads; sync extends the set without authoring merges. The merged fact perspective is held in a **fact cache** that is rebuilt lazily: sync updates the head set without folding policy, and the cache is rebuilt only when a query, session, or local author needs it. Sync bursts that aren't followed by a consumer event collapse into a single rebuild. Peers that have received the same authored commands have equal head sets regardless of arrival order, so cross-peer convergence completes in the rounds needed to propagate authored commands alone — no merge-of-merges cascade.
```
mooreoak: Dose this mean we don't issue effects until once of `query, session, or local` happen. This would keep the app form hearing about changes to the graph until it too an action. 

If I am not misunderstanding this we should have commit compute the cash.

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Problem
+
+`Transaction::commit` collapses every divergent head into a single graph head before returning, even in sync-only transactions. Each merge runs the braid algorithm over the divergent region between the existing graph head and the new commands.
+
+When P peers converge on N total commands, each peer ingests roughly N commands in batches from the other peers. Every batch triggers a braid whose region grows toward N as convergence progresses, and every braid writes a merge segment that subsequent syncs must walk through. Total work is O(P·N²).
```
mooreoak: Why are we calling commit on each batch instead of when the sync session is complete?

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+Lazy merging targets both. Deferring the braid to consumer events and writing no merge segments during sync drops the work to O(P·N) — O(P·N) sync ingestion plus O(N) per-peer rebuild. On the profiled workload that's roughly 4,000 reads instead of 2.5M.
+
+## Key Insight
+
+The multi-head state has two persisted parts:
+
+- **Head set.** Concurrent locations representing the current state, replacing the single `head: Location`.
+- **Fact cache.** The merged fact perspective for some head set, persisted independently of any merge segment. The cache is *fresh* when its fingerprint matches the current head set; otherwise *stale*.
+
+Sync ingestion updates the head set and lets the cache go stale. The cache is rebuilt on demand when a query, session, or author event needs it.
```
mooreoak: We will still need to braid any time we seen a merge as we can't compute the origin perspective without braiding and we can add a command to the graph without it being valid at it's origin perspective. 

This might be implied by one of the comments below but I am unsure.

docs/lazy-merges.md
mooreoak: How could we end up with a new command where one or more of the current heads are it's ansesters but not it's parent?

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+update_head_set(cmd):
+    match cmd.parent():
+        Single(parent):
+            if parent in head_set:
+                replace parent with cmd in head_set
+            else:
+                add cmd to head_set
+                minimize head_set (remove ancestors)
+        Merge(left, right):
+            if {left, right} are both in head_set:
```
mooreoak: what about the case where only either left or right are in the head set?

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+            else:
+                add cmd to head_set
+                minimize head_set
+        None:
+            // init command; head set starts as {cmd}
+```
+
+### Cache read and rebuild
+
+Every cache read goes through `read_cache`, which checks freshness and rebuilds if stale. Query, session, and author callers do not call a separate `ensure_cache_fresh`; they read the cache and the rebuild happens implicitly. Queries pass a `NullSink`; author and the public `refresh_cache` API thread the caller's sink so braid-fold effects flow out during rebuild.
```
mooreoak: Using a null sync here will supres the effects that have not been given to the app yet. I really think we should compute the cash on every commit and just only commit when we are done with all the round of syncing. That will proved more predictable behavior too. Otherwise sometimes doing things like queries will be fast and other times it will cause a braid to happen.

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+- **Many syncs, no local consumer in between.** Eager pays a full braid per sync event and writes a merge segment for each. Lazy pays O(1) per sync; a single rebuild at the next consumer event amortizes the braid work across the entire burst.
+- **Many syncs, frequent queries.** Each query that finds the cache stale triggers a rebuild. In the worst case (one query per sync) lazy and eager do equivalent braid work; lazy still saves merge segment writes and convergence rounds.
+- **One sync, immediate author.** Same braid work either way.
+
+## Edge Cases
+
+### Concurrent transactions
+
+`Transaction::commit` errors with `ClientError::ConcurrentTransaction` when `original_head != storage.get_head()`. In the head-set model the check becomes: every member of `original_heads` must still be present in the current head set or covered by a descendant. Elements removed without a descendant subsuming them reject the transaction.
```
mooreoak: I assume in the current implementation this error causes the new commands (decendents of original_head) to be added to the transaction before commit?

docs/lazy-merges.md
```diff
@@ -0,0 +1,211 @@
...
+### Sessions
+
+`Session::new` reads the fact cache (rebuilding if stale) to seed the session perspective instead of using `storage.get_head()`. Session commit follows the same author-time collapse rules as the main client.
+
+## Changes
+
+### Storage layer
+
+- `Writer::head() -> Location` becomes `Writer::heads() -> HeadSet`.
+- `Writer::commit(head: Location)` becomes `Writer::commit(heads: HeadSet, fact_cache: FactCache)`.
```
mooreoak, line 89: I don't quite understand why we pass the FactCache in here now. Is that just a missing optimization we could have done before?

mooreoak (commented):
This feels like two ideas in one, lazy heads + lazy braiding. I think the lazy heads is good but I don't currently thing the lazy braiding is a good idea as it will delay informing the calling app of effects when it shouldn't (unless I am misunderstanding it) and will cause surprising variations in performance of APIs depending on when the fact cache is clean.

I also think this needs to motivate the fact cache I don't really understand why it is needed vs what we do now.

docs/lazy-merges.md
benz-spideroak: Because doing it as we are now makes the algorithm go from O(n) to O(n^2). You need several million segment reads vs several thousand to sync 450 commands from 9 peers. The way I'm envisioning this is we'd do an entire sync round. Sync with all of our peers until they have nothing new to send and then braid once at the end. Currently performance is terrible with high concurrency because you end up rebraiding the entire section each time you get commands from a new client. 

We can still call refresh_cache after each sync round so we get the effects. As the algorithms are now there's no way to easily catch a new client up if it gets behind or handle high sync concurrency. This allows us to delay braiding and fixes both of those scenarios.

This is conceptually the same thing as allowing transactions to contain commands from multiple syncs.

docs/lazy-merges.md
benz-spideroak: We remove the command that is in the head set and add the merge. That's what the else is saying. There's probably a clearer way to say that.

docs/lazy-merges.md
benz-spideroak: We can either force a merge or error. Probably error since there are a number of other places in the code where we error if there are more than 256 commands at a max cut level in the braid. I believe that's guaranteed to happen if you have more than 256 peers writing at once.

docs/lazy-merges.md
benz-spideroak: I should make this clearer. Really it just add the new command to the head set and removes any ancestors.

docs/lazy-merges.md
benz-spideroak: Yes. I updated the spec to make that clearer.

docs/lazy-merges.md
benz-spideroak: I updated the spec.

docs/lazy-merges.md
benz-spideroak: Currently the facts are stored on merge commands. If we have multiple heads we don't have a merge to store the facts which is why fact cache is created by this spec.

docs/lazy-merges.md
benz-spideroak: I clarified this section.

docs/lazy-merges.md
benz-spideroak: I clarified this. Concurrent transactions are impossible unless you have multiple clients using the same files, but doing that will lead to data corruption anyway.

docs/lazy-merges.md
benz-spideroak: I moved it back to commit

docs/lazy-merges.md
benz-spideroak: I simplified the spec quite a bit. Now it only handles keeping multiple heads open at once. Optionally you can keep a transaction open across syncs to improve performance.

docs/lazy-merges.md
benz-spideroak: I removed this.

docs/lazy-merges.md
```diff
@@ -0,0 +1,117 @@
...
+Commit writes `self.heads` to storage and rebuilds the fact cache for that head set. The pairwise-merge collapse that today's commit performs is gone.
+
+```
+commit(sink):
+    persist any in-flight perspective into self.heads
+    perspective = braid_n_way(self.heads)
+    for each command in braid_order(self.heads):
+        call_rule(command, perspective, sink)  // emits braid-fold effects
+    fact_cache = write_facts(perspective)
+    storage.commit(self.heads, fact_cache)
```
mooreoak, line 47: I assume `sink.commit()` happens here after `storage.commit()`?

docs/lazy-merges.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+The `call_rule` invocations during the cache rebuild are where the application sees effects for the accumulated sync work.
+
+Init creates the graph and seeds storage with the init command directly; it bypasses this flow.
+
+## Action
+
+A new command needs a single parent. `action()` reads `storage.get_heads()`, runs the pairwise-merge loop to collapse the head set to a single head (writing merge segments along the way), gets a linear perspective at the resulting single head, calls the policy's action handler to generate and apply the new command, and commits. The post-action head set is a one-element set.
```
mooreoak, line 56: > the post-action head set is a one-element set.

I am assume we will still need all the appropriate facts on each merge command as that command could later be the site of a branch?

mooreoak: I think this might be a clear case where having bounded n-way merges would be a win over pair merges. If the size current head set is equal to or smaller then the max children in the n-way merge command then you could commit a n-way merge based on the current fact cache instead of recalculating it.  What I don't understand is what pathological cases might exist if there are different n-way merges happening with overlapping heads.

benz-spideroak: > I am assume we will still need all the appropriate facts on each merge command as that command could later be the site of a branch?

Yes. Merges will always have the appropriate facts.

benz-spideroak: > I think this might be a clear case where having bounded n-way merges would be a win over pair merges. If the size current head set is equal to or smaller then the max children in the n-way merge command then you could commit a n-way merge based on the current fact cache instead of recalculating it. What I don't understand is what pathological cases might exist if there are different n-way merges happening with overlapping heads.

That does sound better. I'm just not sure if it's worth doing here. That change touches quite a few pieces of code.

docs/lazy-merges.md
benz-spideroak: Updated.

docs/lazy-merges.md
mooreoak: Ya let's leave it for a later effort.

mooreoak (approved):

docs/lazy-merges.md
```diff
@@ -0,0 +1,117 @@
...
+## Commit
+
+Commit writes `self.heads` to storage and rebuilds the fact cache for that head set. The pairwise-merge collapse that today's commit performs is gone.
+
+```
+commit(sink):
+    persist any in-flight perspective into self.heads
+    perspective = braid_n_way(self.heads)
+    for each command in braid_order(self.heads):
+        call_rule(command, perspective, sink)  // emits braid-fold effects
```
chip-so, line 45: Does this mean you're more likely to see repeated effects from these head commands? e.g. if you have A and B in the HeadSet and you add head C, you will get effects from A, B, and C?

(This is obviously not a violation of any guarantees, I'm just wrapping my head around it)

docs/lazy-merges.md
```diff
@@ -0,0 +1,117 @@
...
+- New persisted record: `FactCache { fact_index: FactIndexRef }`.
+
+### Transaction
+
+- `Transaction::commit` no longer collapses heads. It rebuilds the fact cache n-way (emitting effects to the caller's sink) and persists the head set and cache.
+
+### Action
+
+- `ClientState::action` reads `storage.get_heads()` instead of `storage.get_head()`.
+- The pairwise-merge collapse loop moves from `Transaction::commit` to `action()`. It produces a single head before the policy's action handler runs.
```
chip-so, line 102: Because calling an action requires the HeadSet to be merged, does this mean effects from the commands in HeadSet are emitted to that action's sink?

docs/lazy-merges.md
benz-spideroak: Yes it does. Previously you'd braid the new head and the merged head. Now you'll braid the new head and the other heads. That will sometimes result in more duplicates than before. There are ways to reduce that, but they get pretty complicated. They're probably not worth doing for performance, but may be worth doing if extra effects will cause problems.

docs/lazy-merges.md
benz-spideroak: The way this is currently written we'll pairwise merge the heads. Each of those merges will emit effects. The solution to this is n way merges. Then we can use the fact cache as the cache for the new merge. We had decided to not do that now, but could if the extra effects are a problem.

docs/lazy-merges.md
chip-so: I don't think it should cause _problems_, as we've said that effects may be emitted multiple times. It's an expected difficulty of eventually-consistent systems. Might be worth emphasizing it a little more in the documentation.

chip-so (approved):

PR #202, opened 2026-05-28
Initial draft of module spec

### LINKED ISSUE #696: policy: Specify the policy module format
We have a loose policy module format. Write a spec to formalize the things we want it to include.

chip-so:
Closes aranya-project/aranya-core#696

### DIFF
--- a/docs/policy-module.md
+++ b/docs/policy-module.md
@@ -1,0 +1,92 @@
+---
+layout: page
+title: Policy Module
+permalink: "/policy-module/"
+---
+
+# Policy Module
+
+This document describes version 1 of the policy module. A "policy module" is a compiled and serialized form of a text policy, which can be loaded into the runtime.
+
+The policy module contains both the compiled VM instructions as well as information that lets the runtime validate that the module meets its requirements.
+
+## Overview
+
+The structure of the module has three main sections:
+
+| Section | Description |
+| - | - |
+| Header | Magic, Module Version, and Size |
+| Policy Data | Compiled policy and related information needed by runtime and VM |
+| Checksum | A SHA-2-256 hash over all previous bytes |
+
+Unless otherwise specified, all sizes are in bytes and all integers are big-endian.
+
+## Sections
+
+### Header
+
+| Field | Size | Description |
+| - | - | - |
+| Magic | 8 | Fixed, [0x50, 0x4d, 0x4f, 0x44, 0xcb, 0xb3, 0xcc, 0xbe] "PMOD˳̾" |
+| Version | 4 | Unsigned integer version number of the module format, currently 1 |
+| Size | 4 | Unsigned integer size of the module in bytes, including the header and checksum |
+
+The header is intended to minimally describe any data needed to parse other sections. In particular, storing the size allows the policy module to be easily embedded in places that do not have natural bounds, like a raw flash partition or a network stream.
+
+### Policy Data
+
+The policy data section is encoded using CBOR. The top level is a map with two fields:
+
+| Field | Type | Description |
+| - | - | - |
+| `program` | `ModuleProgram` | Program code and execution information |
+| `contract` | `ModuleContract` | Metadata defining the capabilities and expectations of the module |
+
+#### `ModuleProgram`
+
+`ModuleProgram` describes operational parts of the module. It is a map with these fields:
+
+| Field | Type | Description |
+| - | - | - |
+| `progmem` | array of `Instruction` | "Program Memory" containing all of the instructions for the module |
+| `labels` | map of `Label` -> `usize` | Maps entry point names to addresses in `progmem` |
+| `globals` | map of `Identifier` -> `ConstValue` | globally defined constant values |
+| `codemap` | `Option<CodeMap>` | Contains a copy of the source and maps instruction ranges to source lines |
+
+#### `ModuleContract`
+
+`ModuleContract` describes metadata used for module and execution validation. It is a map with these fields:
+
+| Field | Type | Description |
+| - | - | - |
+| `serial` | 32-bit unsigned integer | A sequential identifier for policies |
+| `actions` | array of `ActionDef` | All defined action signatures |
+| `commands` | array of `CommandDef` | All defined command signatures |
+| `facts` | array of `FactDefinition` | All defined fact signatures |
+| `structs` | map of `Identifier` -> (array of `FieldDefinition`) | All defined struct signatures |
+| `enums` | map of `Identifier` -> (map of `Identifier` -> `i64`) | All defined enum signatures |
+| `ffis` | array of `FfiContract` | FFI schema to compare against VM implementation |
+
+### Checksum
+
+| Field | Size | Description |
+| - | - | - |
+| Checksum | 32 | SHA-2-256 checksum of all prior bytes |
+
+The checksum is a quick data integrity check over the module. It is positioned at the end so that the calculated checksum is ready when a sequential read arrives at it. Because the header gives the module size, we know when the checksummed bytes end and the checksum begins. No stream seeking should be necessary.
+
+## Implementation
+
+Everything in `ModuleProgram` and `ModuleContract` already exists in the V1 module format, except for `serial`. This is mostly a reorganization of those parts.
+
+The serial number should be specified in the module source, in the front matter.
+
+```
+---
+policy-version: 2
+policy-serial: 25
+---
+```
+
+An absent `policy-serial` field is equivalent to 0.
--- a/docs/policy-module.md
+++ b/docs/policy-module.md
@@ -36,37 +36,301 @@ The header is intended to minimally describe any data needed to parse other sect
 
 ### Policy Data
 
-The policy data section is encoded using CBOR. The top level is a map with two fields:
+The policy data section is encoded using [postcard 1.x](https://docs.rs/postcard/latest/postcard/). The top level is a map with two fields:
 
 | Field | Type | Description |
 | - | - | - |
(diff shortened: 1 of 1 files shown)


docs/policy-module.md
```diff
@@ -0,0 +1,92 @@
...
+| - | - | - |
+| Checksum | 32 | SHA-2-256 checksum of all prior bytes |
+
+The checksum is a quick data integrity check over the module. It is positioned at the end so that the calculated checksum is ready when a sequential read arrives at it. Because the header gives the module size, we know when the checksummed bytes end and the checksum begins. No stream seeking should be necessary.
+
+## Implementation
+
+Everything in `ModuleProgram` and `ModuleContract` already exists in the V1 module format, except for `serial`. This is mostly a reorganization of those parts.
+
+The serial number should be specified in the module source, in the front matter.
```
chip-so: There are comments in the source suggesting that the serial number is used to order policies in some important way. So it seemed inappropriate for this to be a hash or randomly generated. I'm not really sure what the intent was here, but I'm leaving it in the hands of the programmer.

docs/policy-module.md
```diff
@@ -0,0 +1,92 @@
...
+| - | - | - |
+| Magic | 8 | Fixed, [0x50, 0x4d, 0x4f, 0x44, 0xcb, 0xb3, 0xcc, 0xbe] "PMOD˳̾" |
+| Version | 4 | Unsigned integer version number of the module format, currently 1 |
+| Size | 4 | Unsigned integer size of the module in bytes, including the header and checksum |
+
+The header is intended to minimally describe any data needed to parse other sections. In particular, storing the size allows the policy module to be easily embedded in places that do not have natural bounds, like a raw flash partition or a network stream.
+
+### Policy Data
+
+The policy data section is encoded using CBOR. The top level is a map with two fields:
```
chip-so: I was already using CBOR in the compiler to produce policy module files. It seems like a better choice to have something broadly documented and implemented, as opposed to something like `rkyv` or `postcard`.

No other part of our code actually decodes the module files, though (the embedded project round-trips them with `rkyv`, which is probably more appropriate for its needs). So this is up for discussion if something else feels more appropriate.

docs/policy-module.md
```diff
@@ -0,0 +1,92 @@
...
+
+Unless otherwise specified, all sizes are in bytes and all integers are big-endian.
+
+## Sections
+
+### Header
+
+| Field | Size | Description |
+| - | - | - |
+| Magic | 8 | Fixed, [0x50, 0x4d, 0x4f, 0x44, 0xcb, 0xb3, 0xcc, 0xbe] "PMOD˳̾" |
```
chip-so, line 31: The last four bytes here are me being a little cheeky. It's a "Modifier Letter Low Ring" U+02F3 followed by a "Combining Vertical Tilde" U+033E. It kind of looks like "SO" written vertically.

docs/policy-module.md
```diff
@@ -0,0 +1,92 @@
...
+The serial number should be specified in the module source, in the front matter.
+
+```
+---
+policy-version: 2
+policy-serial: 25
+---
+```
+
+An absent `policy-serial` field is equivalent to 0.
```
chip-so: I'd also be willing to make this an error. This was just a low-friction decision.

docs/policy-module.md
```diff
@@ -0,0 +1,92 @@
...
+| `contract` | `ModuleContract` | Metadata defining the capabilities and expectations of the module |
+
+#### `ModuleProgram`
+
+`ModuleProgram` describes operational parts of the module. It is a map with these fields:
+
+| Field | Type | Description |
+| - | - | - |
+| `progmem` | array of `Instruction` | "Program Memory" containing all of the instructions for the module |
+| `labels` | map of `Label` -> `usize` | Maps entry point names to addresses in `progmem` |
```
mooreoak: What is the format /encoding of `Label`

docs/policy-module.md
```diff
@@ -0,0 +1,92 @@
...
+| `globals` | map of `Identifier` -> `ConstValue` | globally defined constant values |
```
mooreoak: What is the format of Identifier and ConstValue?

docs/policy-module.md
mooreoak: I think the very early idea is that if you had a merge and each branch had a different policy the policy at the merge would be the one with the larger serial. I don't think we want to do that any more. I think we only want policies to change at finalization. This will mean that there is always a clear policy at any perspective.

docs/policy-module.md
mooreoak: My inclination is that the policy serial should be derivable from the normalized AST of the policy but also I am ok with this for now.

mooreoak (commented):
This looks good my only comment is that there are a lot of value types like `CommandDef` which I think might be rust types? It would be nice if we just included those in the spec.

docs/policy-module.md
chip-so: If it doesn't need to be monotonically increasing, that's fine. But if we want some kind of derived value, it is only a u32, and this should probably specify something larger to avoid collision attacks.

mooreoak: ya if it's derived it should probbly be a sha256 hash over the normalized AST.

chip-so:
> This looks good my only comment is that there are a lot of value types like `CommandDef` which I think might be rust types? It would be nice if we just included those in the spec.

Yes, they're rust types. That's why I asked how detailed it needed to be. I'll flesh them out in the spec, then.

docs/policy-module.md
```diff
@@ -0,0 +1,134 @@
...
+
+`LabelType` describes the type of label and has these string values:
+
+- `Action`
+- `CommandPolicy`
+- `CommandRecall`
+- `CommandSeal`
+- `CommandOpen`
+- `Temporary`
+- `Function`
```
chip-so: This is not complete but I wanted to push what I had to bring up a couple points.

1. I've done some digging into how CBOR/`ciborium` encodes values.
   1. Simple enum variants get encoded as strings but tuple variants get encoded as maps. This is _weird_ when you dissect it and makes the descriptions in this spec strange, but I don't think it actually causes any problems.
   2. CBOR is basically binary JSON, so all identifiers (field names, enums ,etc.) are encoded as text. That _does_ make it agnostic of the underlying enum/field ordering, but it obviously has a space cost. I don't think that's a problem, but I thought it was worth discussing.
2. Diving into all the types here exposes a lot of things I think I'd consider internal or just redundant (e.g. `Target::Unresolved` vs `Target::Resolved` - `Unresolved` shouldn't appear in a compiled program, which makes `Resolved` pointless). To me, this begs that the module format needs a different representation for these types or custom serializer/deserializers, but I don't know if that's effort well-spent.

chip-so: After some discussion:

1. I've decided to use postcard instead. It's well specified and more compact.
2. I've decided to just make public-facing module formats for these types, with conversion to the internal types.

docs/policy-module.md
chip-so: I've swapped out `serial` for a signature.

docs/policy-module.md
```diff
@@ -0,0 +1,360 @@
...
+| `globals` | map of [`Identifier`](#identifier) -> [`ConstValue`](#constvalue) | globally defined constant values |
+| `codemap` | `Option<` [`CodeMap`](#codemap) `>` | Contains a copy of the source and maps instruction ranges to source lines |
+
+#### `Contract`
+
+`Contract` describes metadata used for module and execution validation. It is a map with these fields:
+
+| Field | Type | Description |
+| - | - | - |
+| `signature` | 32 byte array | A SHA-256 hash over the normalized AST |
```
mooreoak, line 63: I think this is fine for version one but we probably want the hash of a lowering as there are things in the AST the provided safety guarantees the do not effect the behavior of the program just which programs will be accepted.  I think it is worth not having these in the signature as it will let us add them without changing the program siginture.

docs/policy-module.md
```diff
@@ -0,0 +1,360 @@
...
+| `name` | [`Identifier`](#identifier) | The name of the struct |
+| `fields` | map of [`Identifier`](#identifier) -> [`ConstValue`](#constvalue) | The fields of the struct |
+
+#### `CodeMap`
+
+`CodeMap` maps between compiled instructions and source code locations. It is a map with these fields:
+
+| Field | Type | Description |
+| - | - | - |
+| `text` | [`Text`](#text) | The policy source |
```
mooreoak, line 194: It might be nice to to define an encoding of the text, not just so we know it's ascii or utf8 or whatever but also so we could add compression. That would allow the CodeMap to be present in more envioments.

docs/policy-module.md
```diff
@@ -0,0 +1,360 @@
...
+A `Field` describes a name and type. It is used in both function parameters and struct fields. It is a map with these fields:
+
+| Field | Type | Description |
+| - | - | - |
+| `name` | [`Identifier`](#identifier) | The name of the field |
+| `ty` | [`TypeKind`](#typekind) | The type of the field |
+
+#### `Meta`
+
+`Meta` describes in more detail actions that the VM takes, so that those actions can be analyzed before execution time. It is an enumeration with these variants:
```
mooreoak, line 316: It's not clear to me what this is from this text.

mooreoak (approved):
There are a couple of changes I would *like* to see but I won't block it on that if you don't want to do them.

docs/policy-module.md
chip-so: The encoding of `Text` is defined in its section.

apetkov-so (approved):

PR #203, opened 2026-05-29
Document `or` operator.

apetkov-so:


policy-book/src/reference/expressions/operators.md
```diff
@@ -47,10 +47,48 @@ Integers can be compared against each other.
...
 runtime exception.
 
+The `or` operator requires A to be `optional[T]` and B to be of type `T`.
+The result type is `T`. B is only evaluated when A is None (short-circuit
+evaluation). `or` is right-associative, so `a or b or c` is parsed as
+`a or (b or c)`.
+
+```
+let x = Some(1)
+let y = None
```
chip-so: `x` and `y` aren't used in this example.

policy-book/src/reference/expressions/operators.md
```diff
@@ -47,10 +47,48 @@ Integers can be compared against each other.
...
+
+let a = Some(1) or 0    // a = 1
+let b = None or 0       // b = 0
+
+// chaining: evaluates left to right, stopping at the first Some.
```
chip-so: This seems to confuse the earlier statement about evaluation being right-associative. It should evaluate `Some(1) or 42` to `1`, then `None or 1` to `1`, right?

apetkov-so: Yes. (I wrote a test to prove it) So the logic is correct, but the wording could be clearer.
```
None or Some(1) or 42
None or (1 or 42)
None or 1
1
```

jdygert-spok: It has right associativity but does left-to-right evaluation with short circuiting. So it does group as `None or (Some(1) or 42)`. But it first evaluates the `None`, then falls back to evaluate the right side group. So then it evaluates `Some(1)` and then short circuits to `1` and doesn't need to evaluate 42.

Think of `a or b or c` as:

```
if a is Some then unwrap a else (if b is Some then unwrap b else c)
```

policy-book/src/reference/expressions/operators.md
```diff
@@ -52,21 +52,24 @@ Integers can be compared against each other.
...
-evaluation). `or` is right-associative, so `a or b or c` is parsed as
-`a or (b or c)`.
+### Coalescing `or`
 
+The `or` operator requires the LHS to be `optional[T]` and the RHS to be of type `T`.
+The result type is `T`. The RHS is only evaluated when the LHS is None (short-circuit evaluation).
+
+```policy
+let a = Some(5) or foo() // c = 5, foo() not called
+let b = None or Some(0)  // b = 0
```
chip-so: This should be `b = Some(0)`, yeah? Maybe you meant to just put a `0` on the right side?

policy-book/src/reference/expressions/operators.md
```diff
@@ -47,10 +47,51 @@ Integers can be compared against each other.
...
+```
+
+#### Associativity and evaluation order
+
+`or` is right-associative, so `a or b or c` is parsed as `a or (b or c)`. Evaluation order is
+left-to-right, meaning if both sides are non-None, the left side is chosen.
+
+```policy
+let one = None or Some(1) or 42 
+// None or (Some(1) or 42) -> None or 1 -> 1
```
jdygert-spok: Should it be this? 

```
None or (Some(1) or 42) -> Some(1) or 42 -> 1
```

policy-book/src/reference/expressions/operators.md
apetkov-so: Ah yes, the RHS is not unwrapped. I was thinking whether it would be a good idea to unwrap the RHS (for consistency and to avoid having to unwrap afterward)... But you wouldn't normally write expressions where the final default is optional - unless that was the intent.

`let f = query Foo [] or Foo {} // f is Foo`

policy-book/src/reference/expressions/operators.md
apetkov-so: Well, the parenthesized expression is evaluated first, so `None` remains until after. The parens resolve to `1`, so we're left with `None or 1`.

apetkov-so: Actually, you're right - it's evaluated strictly left-to-right, not algebraically.

chip-so (approved):

apetkov-so:
The remaining failing tests are actually a bit tricky to fix. The issue is that we don't set the context before calling a recall, so we fail to set the effect `recalled` field. The way it used to work was, the runtime called `RunState::set_context` explicitly before running the recall block. Actually all contexts (open, seal, recall) were run in distinct RunState instances, so we never had to switch contexts during the same RunState execution. But now that we do change contexts - like transitioning from policy to recall - we need a mechanism, e.g. new instructions to push/pop the context.

apetkov-so:
The `or` operator was documented as part of a different ticket.

PR #204, opened 2026-06-03
200 action return values spec

### LINKED ISSUE #200: Action return values spec
Now that `check` no longer exits, actions need an alternative way to early-terminate, and communicate the reason to the runtime. The consensus is that actions should return `result`, but the details need to be fleshed out.

apetkov-so:


docs/action-result-spec.md
```diff
@@ -0,0 +1,69 @@
...
+
+1. Actions gain *optional* `result` return type
+
+    ```policy
+    action infallible() {}
+    action can_fail() result[unit, enum Error] {}
+    ```
+    
+    The success type is limited to `unit`; returning success values from actions is disallowed
+    because the action result is not authoritative, the sink is.
```
chip-so: I think "authoritative" here means "authoritative in terms of graph update success". I think it's worth spelling that out.

docs/action-result-spec.md
```diff
@@ -0,0 +1,69 @@
...
+with the payload. The `PolicyError::Check` variant gains a payload of type `Value`.
+- If the value is `unit`, the action succeeded. If it is not found, the action is infallible. In
+either case, return a success value to the caller (client/session).
+
+
+
+
+TODO:
+- if PolicyError::Check has required value, how does that work with #648 (invoke-recall-from-policy)?
+    actions return Err(<value>), but commands don't. yet both produce Check exit
```
chip-so: I think you already solved this by making the value optional? Or I'm not understanding the problem.

chip-so (approved):

apetkov-so:
Already implemented, and documented.

PR #205, opened 2026-06-12
Sync metrics

benz-spideroak:
I don't know if we actually want to commit this. @mooreoak asked me to put it together to be able to share with customers.


chip-so:
IMO it should only be merged if it's something we want to maintain as a part of the project. As a one-off report, I'd leave it out.

PR #206, opened 2026-07-29
Document recent language changes

apetkov-so:
Document recent language changes: action results, coalescing or, removal of unwrap operators.

### DIFF
--- a/docs/policy-v2.md
+++ b/docs/policy-v2.md
@@ -311,8 +311,8 @@ return value as evidence that its commands took effect.
 
 An action with a return type must return on every path, using
 `return Ok(Unit)` for success and `return Err(e)` for failure. Returning
-`Err` terminates the policy with a rejection, meaning that none of the
-commands it published are accepted onto the graph.
+`Err` terminates the action, meaning that none of the commands it published
+are accepted onto the graph.
 
 An action with no declared return type cannot report a failure of its
 own. Such actions still fail if any published command fails.
--- a/docs/policy-v2.md
+++ b/docs/policy-v2.md
@@ -282,3 +282,56 @@ action frob(x Bar) {
     let f = x substruct Foo
 }
 ```
+
+## Return from Actions
+
+Actions can declare a return type, so that an action can report
+failure to the calling application.
+
+```policy
+enum Error { InvalidAmount, NoSuchDevice }
+
+action deposit(device id, amount int) result[unit, enum Error] {
+    check amount > 0 else return Err(Error::InvalidAmount)
+    let account = query Account[device: device] or return Err(Error::NoSuchDevice)
+    publish Deposit { device: device, amount: amount }
+    return Ok(Unit)
+}
+```
+
+The return type is optional. If specified, it must be a
+`result[unit, E]`, where `E` may be any policy type. It is commonly an
+`enum`, which gives the application a fixed set of failure cases to
+match on.
+
+The success type is limited to `unit` (whose only value is `Unit`). An
+action succeeding means only that the action itself ran to completion.
+It does not mean the commands it published were accepted onto the graph. Limiting the success type keeps applications from treating an action's
+return value as evidence that its commands took effect.
+
+An action with a return type must return on every path, using
+`return Ok(Unit)` for success and `return Err(e)` for failure. Returning
+`Err` terminates the policy with a rejection, meaning that none of the
+commands it published are accepted onto the graph.
+
+An action with no declared return type cannot report a failure of its
+own. Such actions still fail if any published command fails.
+
+The `return` statement is otherwise unchanged, and remains invalid in
+`policy` and `recall` blocks.
+
+## Errors and control flow
+
+Error handling within policy code is described in [Policy Lang: "or" and
+"recall"](/docs/policy-lang-or-recall.md). In summary:
+
+- `or` is an optional coalescing operator. `a or b` is the value inside
+  `a` when it is `Some`, and `b` otherwise.
+- `recall` is a statement and expression that transfers evaluation to a
+  named recall block. Recall blocks are now named and take parameters,
+  and a command may define more than one.
+- `check` now requires an `else` clause holding a terminal expression
+  (`return`, `recall`, or `todo()`), which determines how the failed
+  check is reported.
+- `unwrap` and `check_unwrap` are removed. Their uses are covered by
+  `or` with a fallback value, a `return`, a `recall`, or `todo()`.
--- a/docs/quic-channels.md
+++ b/docs/quic-channels.md
@@ -653,61 +653,73 @@ function is_valid_psk_length(size int) bool {
 
 // Returns the channel operations that this device is allowed to
 // perform for a particular label.
-function get_allowed_op(device_id id, label_id id) enum ChanOp {
-    let assigned = check_unwrap query AssignedLabel[device_id: device_id, label_id: label_id]
-    return assigned.op
+function get_allowed_op(device_id id, label_id id) result[enum ChanOp, int] {
+    let assigned = query AssignedLabel[device_id: device_id, label_id: label_id] or return Err(0)
+    return Ok(assigned.op)
 }
 
 // Reports whether the devices have permission to create
 // a bidirectional AQC channel with each other.
-function can_create_aqc_bidi_channel(device1 id, device2 id, label_id id) bool {
+function can_create_aqc_bidi_channel(device1 id, device2 id, label_id id) result[bool, int] {
     // Devices cannot create channels with themselves.
     //
     // This should have been caught by the AQC FFI, so check
     // instead of just returning false.
-    check device1 != device2
+    check device1 != device2 else return Err(0)
 
     // Both devices must have permissions to read (recev) and
     // write (send) data.
-    let device1_op = get_allowed_op(device1, label_id)
-    if device1_op != ChanOp::SendRecv {
-        return false
+    let device1_op = match get_allowed_op(device1, label_id) {
+        Ok(id) => {
+            if device1_op != ChanOp::SendRecv {
+                return Ok(false)
+            }
+        }
+        Err(e) => return Err(e)
     }
 
-    let device2_op = get_allowed_op(device2, label_id)
-    if device2_op != ChanOp::SendRecv {
-        return false
+    let device2_op = match get_allowed_op(device2, label_id) {
+        Ok(id) => {
+            if device2_op != ChanOp::SendRecv {
+                return Ok(false)
+            }
+        }
+        Err(e) => return Err(e)
     }
 
-    return true
+    return Ok(true)
 }
 
 // Reports whether the devices have permission to create
 // a unidirectional AQC channel with each other.
-function can_create_aqc_uni_channel(sender_id id, receiver_id id, label_id id) bool {
+function can_create_aqc_uni_channel(sender_id id, receiver_id id, label_id id) result[bool, int] {
     // Devices cannot create channels with themselves.
     //
     // This should have been caught by the AQC FFI, so check
     // instead of just returning false.
-    check sender_id != receiver_id
+    check sender_id != receiver_id else return Err(0)
 
     // The writer must have permissions to write (send) data.
-    let writer_op = get_allowed_op(sender_id, label_id)
-    match writer_op {
-        ChanOp::RecvOnly => { return false }
-        ChanOp::SendOnly => {}
-        ChanOp::SendRecv => {}
+    let writer_op = match get_allowed_op(sender_id, label_id) {
+        Ok(op) => {
+            if writer_op == ChanOp::RecvOnly {
+                return Ok(false)
+            }
+        }
+        Err(e) => return Err(e)
     }
 
     // The reader must have permission to read (receive) data.
-    let reader_op = get_allowed_op(receiver_id, label_id)
-    match reader_op {
-        ChanOp::RecvOnly => {}
-        ChanOp::SendOnly => { return false }
-        ChanOp::SendRecv => {}
+    let reader_op = match get_allowed_op(receiver_id, label_id) {
+        Ok(op) => {
+            if reader_op == ChanOp::SendOnly {
+                return Ok(false)
+            }
+        }
+        Err(e) => return Err(e)
     }
 
-    return true
+    return Ok(true)
 }
 ```
 
@@ -826,16 +838,16 @@ command AqcCreateBidiChannel {
 
     policy {
         let author = get_valid_device(envelope::author_id(envelope))
-        let peer = check_unwrap find_existing_device(this.peer_id)
+        let peer = find_existing_device(this.peer_id) or recall unauthorized()
 
-        check is_valid_psk_length(this.psk_length_in_bytes)
+        check is_valid_psk_length(this.psk_length_in_bytes) else recall unauthorized()
 
         // The label must exist.
-        let label = check_unwrap query Label[label_id: this.label_id]
+        let label = query Label[label_id: this.label_id] or recall unauthorized()
 
         // Check that both devices are allowed to participate in
         // this bidirectional channel.
-        check can_create_aqc_bidi_channel(author.device_id, peer.device_id, label.label_id)
+        check can_create_aqc_bidi_channel(author.device_id, peer.device_id, label.label_id) else recall unauthorized()
 
         // NB: Check roles, other ACLs here.
 
@@ -879,8 +891,12 @@ command AqcCreateBidiChannel {
         } else {
             // This is an off-graph session command, so only the
             // communicating peers should process this command.
-            check false
+            check false else recall unauthorized()
         }
+    }
+
+    recall unauthorized() {
+        finish {}
     }
 }
 ```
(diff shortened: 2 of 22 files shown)


docs/quic-channels.md
```diff
@@ -653,61 +653,73 @@ function is_valid_psk_length(size int) bool {
...
-    let device1_op = get_allowed_op(device1, label_id)
-    if device1_op != ChanOp::SendRecv {
-        return false
+    let device1_op = match get_allowed_op(device1, label_id) {
+        Ok(id) => {
+            if device1_op != ChanOp::SendRecv {
+                return Ok(false)
+            }
+        }
+        Err(e) => return Err(e)
```
chip-so: This looks erroneous. Somewhere in here something should be producing a value, but the only non-terminal branch just ends on line 676. I'm guessing this was supposed to be a block expression ending with `: id`.

And there are several other instances of this below.

policy-book/src/reference/statements/check.md
```diff
@@ -14,36 +14,42 @@
...
-take corrective measures after a command is no longer valid. This could
-for example, cascade to deleting a Fact or emitting an Effect that the
-application can use to take further action.
+permissions, `check` is the appropriate tool to enforce that. The `else`
+expression decides what happens on failure:
 
-```
+- In a command `policy` block, `recall` the command so the policy can
+  take corrective measures after a command is no longer valid. This
+  could, for example, cascade to deleting a fact or emitting an effect
```
chip-so: capitalize `Fact` and `Effect`.

policy-book/src/reference/statements/map.md
```diff
@@ -30,15 +30,17 @@ facts. The name given after `as` is scoped to the block.
...
  ```
 
 Like `query` and related functions, fact values or the entire value part
 of the fact literal can be omitted. And likewise, bind values must
 [follow the positioning
-rules](../expressions/functions/queries.md#bind-marker).
\ No newline at end of file
+rules](../expressions/functions/queries.md#bind-marker).
+
+The map statement is allowed only in actions, because the main use case is to publish commands for each fact; only actions may publish.
```
chip-so, line 48: This is true, but doesn't really describe the underlying reasoning - `map` is allowed in actions because they are run once, on the client that creates commands. We prohibit them in `policy` because it can be re-evaluated many times and loops can cause an explosion in run-time cost.

policy-book/src/reference/top-level/actions.md
```diff
@@ -1,21 +1,43 @@
 # Actions
 
-```
+```policy
```
chip-so, line 3: This is also happening throughout. It's not exactly _wrong_, but it's also entirely useless for a document that isn't actually a policy.

policy-book/src/reference/top-level/actions.md
```diff
@@ -1,21 +1,43 @@
...
 action foo(a int, b string) {
     let cmd = Foo{a: a + 1, b: b}
     publish cmd
 }
 ```
 
+```policy
+let E_INVALID_ID = -1
```
chip-so: I'd like to encourage using enums instead of integers in examples. This functions fine, but enums provide a clearer and safer value (which is also more obvious in things like generated policy interface code).

policy-book/src/reference/types/optional.md
```diff
@@ -1,26 +1,33 @@
 # Optional Types
 
 A type which can contain a value (`Some`) or no value (`None`). The type
-of the value is specified after `optional`, e.g. `optional int`. The
+of the value is specified after `optional`, e.g. `optional[int]`. The
```
chip-so: The syntax with brackets is `option[T]`, not `optional[T]`.

policy-book/src/reference/types/optional.md
```diff
@@ -1,26 +1,33 @@
...
 // declare an optional field
 effect Foo {
-     a string,
-     b optional int,
+    a string,
+    b optional[int]
 }
 
 // initialize an optional
-// type of these player variables is `optional string`
```
chip-so, line 15: Not sure why this was removed.

policy-book/src/reference/errors.md
```diff
@@ -62,8 +61,10 @@ command FailCommand {
     // omit seal and open for example
 
     policy {
-        check !this.fail
+        check this.fail else recall failed()
```
chip-so: Removing `!` inverts the meaning here.

policy-book/src/reference/top-level/actions.md
apetkov-so: It adds syntax highlighting. :)

chip-so: Oh, did we add syntax highlighting for policy lang? I missed that.

policy-book/src/reference/types/optional.md
apetkov-so: I thought that was a bit unnecessary, but okay.

However, the second var is of type `option[never]`, but we can't really write that because the never type is not exposed in the policy grammar.

chip-so: It's meant to demonstrate the concept in detail. I don't think any point can be too obvious here.

And yeah, you are correct about the second type. Maybe it would be clearer for the second one to explain that a literal `None` has a type that matches any optional? That kind of explains the concept without using the explicit internal type signature.

policy-book/src/reference/top-level/actions.md
apetkov-so: Well, it only works in the editor, not in preview. I've read that there are extensions that can handle custom language syntax rules, but I haven't been able to get any to work.

chip-so: Ah. Well, fair enough. If it's useful during editing, let's keep it.

chip-so (dismissed):

jdygert-spok:
Should we document these in `policy-v2.md` as well?

policy-book/src/reference/errors.md
```diff
@@ -1,32 +1,31 @@
...
+- Runtime exceptions, caused by code violating some execution invariant.
 
 ## Check failures
 
-The `check` statement and the `check_unwrap` expression report failure
-by exiting with a _check failure_. A check failure is distinct from
-other errors in that it causes execution to fall to the `recall` block.
+The `check` statement either returns from the current function/action or
+recalls the current command by executing a named recall block. The recall
+block terminates with a _check failure_.
```
jdygert-spok: Are we switching to calling this a rejection?

policy-book/src/reference/expressions/block-expressions.md
```diff
@@ -35,8 +35,8 @@ Or [`match` expressions](functions/if-match.md#match).
 action foo(location id, cap enum Capability) {
     let level = match x {
         Capability::Jump => {
-            let user_perms = unwrap_check
-                query UserJump[location: location]=>{allowed_level: ?}
+            let user_perms =
+                query UserJump[location: location]=>{allowed_level: ?} or return
```
jdygert-spok: This isn't the point of this example but bare return isn't allowed. Let's just replace with a function call to not confuse.

apetkov-so:
> Should we document these in `policy-v2.md` as well?

I was thinking of merging the v1 and v2 docs into a single policy.md file. We don't support multiple versions, so there's little value to separating the docs by version.

policy-book/src/reference/expressions/block-expressions.md
apetkov-so: Ah yes. I need to update the actions.md documentation to remove the part about early returning.

chip-so:
> > Should we document these in `policy-v2.md` as well?
> 
> I was thinking of merging the v1 and v2 docs into a single policy.md file. We don't support multiple versions, so there's little value to separating the docs by version.

I'd like to preserve the existing specs as is (and somehow mark them as deprecated or superseded), but if you want to add a cohesive "final" v2 spec I think that sounds fine. The policy book is intended to be the "living" and continually updated reference for the language (which reminds me, the changes in this PR should be described in a new revision in `policy-book/src/revision-history.md`).

policy-book/src/reference/expressions/operators.md
```diff
@@ -57,7 +56,7 @@ runtime exception.
 |----------|----|
 | 1        | `.` |
 | 2        | `substruct`, `as` |
-| 3        | `-` (prefix), `!`, `unwrap`, `check_unwrap` |
+| 3        | `-` (prefix), `!` |
```
jdygert-spok, line 60: Missing `or` in the precedence table.

docs/policy-v2.md
```diff
@@ -282,3 +282,56 @@ action frob(x Bar) {
...
+match on.
+
+The success type is limited to `unit` (whose only value is `Unit`). An
+action succeeding means only that the action itself ran to completion.
+It does not mean the commands it published were accepted onto the graph. Limiting the success type keeps applications from treating an action's
+return value as evidence that its commands took effect.
+
+An action with a return type must return on every path, using
+`return Ok(Unit)` for success and `return Err(e)` for failure. Returning
+`Err` terminates the policy with a rejection, meaning that none of the
```
chip-so: Would probably be more clear to say "terminates the action", to distinguish from policy failure.

policy-book/src/concepts/facts.md
```diff
@@ -88,8 +88,8 @@ we fetch this fact.
 
 We use [`or`](../reference/expressions/operators.md#optional-operators)
 with `recall` to unwrap the optional given to us by the query. If the
-optional is `None`, the named `recall` block runs, producing a [check
-failure](../reference/errors.md#check-failures). `check` and `or ...
+optional is `None`, the named `recall` block runs, producing a [rejection
+failure](../reference/errors.md#rejection). `check` and `or ...
```
chip-so: I would just call it a "rejection", not a "rejection failure".

chip-so (dismissed):
This looks good. Just a couple of comments on clarity of wording.

policy-book/src/reference/errors.md
apetkov-so: Updated.

policy-book/src/reference/types/optional.md
```diff
@@ -1,26 +1,83 @@
 # Optional Types
 
 A type which can contain a value (`Some`) or no value (`None`). The type
-of the value is specified after `optional`, e.g. `optional int`. The
-literal expressions for optionals are `None` and `Some(❮expression❯)`.
+of the value is specified after `option`, e.g. `option[int]`. The literal
```
chip-so: I think this should probably be reworded given the new syntax. "The type of the value is specified in brackets, e.g. `option[int]`."

policy-book/src/reference/errors.md
```diff
@@ -1,32 +1,47 @@
...
+- Command recalls, usually caused by not meeting the expectations of a
+  `check` statement.
+- Runtime exceptions, caused by code violating some execution invariant.
 
-The `check` statement and the `check_unwrap` expression report failure
-by exiting with a _check failure_. A check failure is distinct from
-other errors in that it causes execution to fall to the `recall` block.
-A check failure represents a failed precondition that the policy author
-recognized could be possible in normal operation.
+The difference is that a rejection is an outcome the policy author
```
chip-so, line 9: This sentence talks about "rejection" but nothing above defines the term. I assume this was supposed to be described in the first point, or this sentence meant "recalls" and rejection is described below.

chip-so (commented):
In addition to the below, we need to update the allowed types table in `src/reference/top-level/facts.md` to include enums and unit. And we should probably explain the unit type in the Types section.

docs/policy-v2.md
```diff
@@ -282,3 +282,56 @@ action frob(x Bar) {
...
+```
+
+The return type is optional. If specified, it must be a
+`result[unit, E]`, where `E` may be any policy type. It is commonly an
+`enum`, which gives the application a fixed set of failure cases to
+match on.
+
+The success type is limited to `unit` (whose only value is `Unit`). An
+action succeeding means only that the action itself ran to completion.
+It does not mean the commands it published were accepted onto the graph. Limiting the success type keeps applications from treating an action's
```
jdygert-spok: I don't think this is quite the right message since an action's success _does_ mean that its commands are accepted onto the graph, even if they might be rejected later. I know @mooreoak had some reasoning against action return values but I don't remember it exactly.

chip-so (approved):

docs/policy-v2.md
apetkov-so: I think it's because the commands could be rejected later. Apps should not assume that a successful action result means its commands were accepted.

mooreoak: @jdygert-spok yes the transaction will fail if that command is not accepted at it's origin location. So there will be an error of some kind if the command was not added to the graph. 

The reason for not having return values is that it creates two result paths, the return value and the sink. All  devices that sync the changes get it in the sink so it would be strange to get something extra in the action return. I also worry that this could cause application state above Aranya to diverge depending on if the device is the one that called the action vs on that received the action effects thought a sync.

policy-book/src/reference/statements/map.md
mooreoak: To put a tighter point our your description Chip: Map is not aloud in policy because it offers a way to unbounded execution. 

(In the future we could add an approach to add bounded map if we thought that would be useful)

policy-book/src/reference/errors.md
```diff
@@ -37,23 +52,44 @@ instead return an error to the application.
...
+action foo(a int) result[unit, enum Error] {
+    check a > 0 else return Err(Error::InvalidID)
+    publish Foo{a: a}
+    return Ok(Unit)
+}
 ```
-action do_nothing() {
+
+An action can fail as you'd expect, but it can also fail if its
+`publish`ed commands fail. Regardless of whether the commands
```
mooreoak: I would say this "but it can also fail if its `publish`ed commands fail." as "but it can also fail if its
`publish`ed commands are recalled at their origin perspective ."

mooreoak (changes requested):
I noticed we updated the docs for quic-channels.md. I don't think we have that feature any more. Is that misnamed or should it just be deleted?

apetkov-so:
> I noticed we updated the docs for quic-channels.md. I don't think we have that feature any more. Is that misnamed or should it just be deleted?

Claude updated it to fix policy syntax. 

> I noticed we updated the docs for quic-channels.md. I don't think we have that feature any more. Is that misnamed or should it just be deleted?

Deleted.

docs/policy-v2.md
```diff
@@ -282,3 +282,57 @@ action frob(x Bar) {
...
+- `or` is an optional coalescing operator. `a or b` is the value inside
+  `a` when it is `Some`, and `b` otherwise.
+- `recall` is a statement and expression that transfers evaluation to a
+  named recall block. Recall blocks are now named and take parameters,
+  and a command may define more than one.
+- `check` now requires an `else` clause holding a terminal expression
+  (`return`, `recall`, or `todo()`), which determines how the failed
+  check is reported.
+- `unwrap` and `check_unwrap` are removed. Their uses are covered by
+  `or` with a fallback value, a `return`, a `recall`, or `todo()`.
```
jdygert-spok: Also covered by binding match.

policy-book/src/reference/expressions/functions/todo.md
```diff
@@ -12,4 +12,17 @@ action foo(type enum Bar) {
...
+`unwrap` operator the language no longer has:
+
+```policy
+let x = y or todo()
+```
+
+Handling the `None` with a fallback value, a `return`, or a `recall` is
+always preferable; `todo()` states that the case is simply not handled
+yet. To state instead that a case should never be reached, see
+[`test_fail()`](test-fail.md).
```
jdygert-spok: This recommends `test_fail` a bit too much, given that our downstream users are expected not to use it.

policy-book/src/reference/statements/check.md
```diff
@@ -14,36 +14,42 @@
...
+check envelope::author_id(envelope) == device else recall unauthorized()
 ```
 
-`check` evaluates a boolean expression and terminates policy execution
-with a check failure if it evaluates to false.
+`check` evaluates a boolean expression. If it is true, execution
+continues with the next statement. If it is false, the `else` expression
+is evaluated instead. The `else` expression must be terminal &ndash; a
+[`recall`](../top-level/commands.md#recall-block), a [`return`](return.md),
+or [`todo()`](../expressions/functions/todo.md) &ndash; so a failed
```
jdygert-spok: Nit: these should be emdashes with no space around them, or personally I'd prefer parentheses.

docs/policy-v2.md
jdygert-spok: The updated text seems reasonable. There's a similar description in both actions.md and errors.md which should match.

policy-book/src/reference/top-level/facts.md
```diff
@@ -21,7 +21,9 @@ storage restrictions. This table defines which types are allowable.
 | `bytes`    | yes      | yes        |
 | `bool`     | yes      | yes        |
 | `id`       | yes      | yes        |
-| `optional` | no       | yes        |
+| `enum`     | yes      | yes        |
+| `option`   | no       | yes        |
+| `result`   | no       | no         |
```
jdygert-spok: Results can be used fact values. There's currently nothing that can't.

jdygert-spok (commented):
Did a full review this time, looking mostly good.

jdygert-spok (approved):

PR #208, opened 2026-07-29
Try operator spec

apetkov-so:


mooreoak (dismissed):

docs/policy-lang-try-operator.md
```diff
@@ -0,0 +1,145 @@
...
+
+## 2. Motivation
+
+Propagation is written by hand today:
+
+```policy
+action foo() result[unit, enum Error] {
+    let value = match try_get_value() {
+        Ok(n) => n
+        _ => return Err(Error::Fail)
```
chip-so: Wouldn't a more analogous match statement be:
```
let value = match try_get_value() {
    Ok(n) => n
    Err(e) => return Err(e)
}
```
?

docs/policy-lang-try-operator.md
```diff
@@ -0,0 +1,145 @@
...
+action foo() result[unit, enum Error] {
+    let value = try_return(succeed)?
+    ...
+}
+```
+
+## 3. Non-goals
+
+- `?` on anything but a call — not on locals, `Ok`/`Err` literals, or field
+  accesses (§4)
```
chip-so: As an operator, IMO the context should be irrelevant. I don't see a reason why it shouldn't work on any result value.

docs/policy-lang-try-operator.md
```diff
@@ -0,0 +1,145 @@
...
+- `?` on `option[T]` - we can use the `or` operator for handling `None`
+- error conversion — the language has no trait system, so no `From` analogue
+
+## 4. Syntax
+
+`?` is a postfix operator whose operand must be a call — it goes directly after
+one, and nowhere else. It is part of the call form rather than a free-standing
+postfix operator, so `x?` on a local, `this.field?`, and `Err(e)?` are parse
+errors. It may repeat: `f(x)??` unwraps a `result[result[T, E], E]`,
+since the chain still starts at a call.
```
chip-so: As an operator, IMO the context should be irrelevant. I don't see a reason why it shouldn't work on any result value.

docs/policy-lang-try-operator.md
```diff
@@ -0,0 +1,145 @@
...
+
+The restriction keeps the parser simple, and avoids ambiguity against the *bind*
+token, which is also `?`. A bind never follows a call, so
+`query Stuff[x: k]=>{y: ?}` is unambiguous, and is easy to parse.
```
chip-so: Bind can only show up alone in that one place, anyway. It doesn't exist in the expression syntax, so there is no ambiguity to resolve.

docs/policy-lang-try-operator.md
```diff
@@ -0,0 +1,145 @@
...
+succeeded. Tracked separately in
+[#800](https://github.com/aranya-project/aranya-core/issues/800).
+
+**Early exit from `map`.** A `map` body's query iterator is only released when
+it is exhausted, so any early exit — an existing `return`, or a new `?` — leaks
+one for the life of the run. This is an existing issue.
+
+**`?` on `option[T]`.** Rust's `?` works on `Option`, so users may expect the
+same here. It stays a compile error: `None` has no error value to propagate, and `or`
+already covers the case. The §5.3 note points at `or` for that reason.
```
chip-so: What the heck is `§5.3`? There are no subsections in this document.

docs/policy-lang-try-operator.md
```diff
@@ -0,0 +1,145 @@
...
+`?` is not valid in a finish block
+```
+
+A `?` on a non-call operand is a parse error, not a type error, and should say
+so plainly: ``the `?` operator can only follow a call``.
+
+## 6. Dynamic semantics
+
+`f(a)?` is an expression of type `T`, so it goes anywhere an expression of that
+type goes: `let x = f(a)?`, a call argument, a struct field, either side of `+`.
```
chip-so: I'm not sure what's "dynamic" about this, but this seems like a restatement of behavior that is already well-specified.

docs/policy-lang-try-operator.md
apetkov-so: Yeah, that is more clear.

docs/policy-lang-try-operator.md
apetkov-so: Removed.

docs/policy-lang-try-operator.md
apetkov-so: I missed that when redacting. Also, the whole Tests section (that also mentions §5.3) is probably unnecessary.

docs/policy-lang-try-operator.md
apetkov-so: Removed.

mooreoak (approved):

